PR #1, opened 2018-07-07
Set up the form for testing purposes

Samudraneel:
Bare set up containing two form fields which will be hooked up with firebase.


src/App.js
```diff
@@ -1,18 +1,24 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
+import FormSubmit from './form/form-submit'
 import './App.css';
 
 class App extends Component {
```
mark-azer: Since this component is stateless and only calls the render function, you can use an arrow function instead of a class.
This should look something like one of the two options:
```
const App = () => {
    return (
        ...
    )
}
```
```
const App = () => (
    ...
)
```

src/form/form-submit.js
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+
+class FormSubmit extends React.Component {
+  constructor(props) {
```
mark-azer: Traditional constructors aren't needed here and if you use arrow functions for your methods, it will bind them to this automatically. A more concise way to write this would be like:

```
state = {
    ...
}

handleChange = (event) => { ... }
```

src/form/form-submit.js
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
+
+class FormSubmit extends React.Component {
```
mark-azer: SInce you imported Component, just use Component instead of React.Component

PR #7, opened 2018-09-30
Adds react-router routes and Header component

livictoria:
- [x] Adds basic react-routers routes for the web application (both private and public routes)
- [x] Adds the header component 
- [x] Adds a global custom `MUITheme` with specific font families and palette colour themes 
- [x] Adds section titles (performances, events, critiques)

image


livictoria:
@Samudraneel yes I agree there is inconsistency, my apologies! IMO, I would prefer having it in one directory/file. The codebase initially has some components with specific style files in the specific folder, I can definitely create a new PR that fixes all these as well as ensures consistent naming convention which we can discuss during a Thursday meeting. Just wanted to get this out and running.

mark-azer:
Ok I took a look at this code and everything seems to be fine. But I agree with Sam's suggestion that in the future we'll open a branch to clean up the inconsistency of where we define styles in the code.

PR #23, opened 2018-12-29
Dynamically load updates to Performances from Firestore

YangStephen:
- Removes a bunch of unnecessary console.logs
- converts us to use snapshots for Performances only right now. The logic here reset the array upon snapshot updates.


src/components/interface/SectionHeader.jsx
```diff
@@ -12,7 +12,8 @@ const SectionHeader = ({
   classes,
   title,
   showNew,
-  showWinner
+  showWinner,
```
livictoria: we don't need the showWinner component anymore so we can remove any logic that has to do with showWinner

src/components/interface/SectionHeader.jsx
```diff
@@ -38,7 +39,8 @@ SectionHeader.propTypes = {
   classes: PropTypes.string.isRequired,
   showWinner: PropTypes.bool,
   showNew: PropTypes.bool,
-  title: PropTypes.string.isRequired
+  title: PropTypes.string.isRequired,
+  eventId: PropTypes.string,
```
livictoria: alphabetical order for `eventId`? :D

src/components/performances/PerformancesSection.jsx
```diff
@@ -43,7 +44,7 @@ class PerformancesSection extends React.Component {
     const collectionName = `events/${eventId}/performances`;
     return (
       <React.Fragment>
-        <SectionHeader title="performance" />
+        <SectionHeader title="performance" eventId={eventId} />
```
livictoria: nitpick - swap `title` and `eventId` ?

YangStephen:
[Adding dynamically updating tables](https://trello.com/c/w8CawUiW/10-adding-dynamically-updating-tables)

src/components/interface/SectionHeader.jsx
livictoria: also can we add an indent for material ui imports

PR #25, opened 2018-12-29
Update EditDropdown to delete performances and adjudications

livictoria:
* delete a performance from `editDropdown`
* delete an adjudication from `editDropdown`

### DIFF
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -49,6 +49,7 @@ class AdjudicationDialog extends React.Component {
         <div style={{ display: 'flex', flexFlow: 'column', margin: '35px' }}>
           <div style={{ display: 'flex', justifyContent: 'space-around', borderBottom: '1px solid #dcdcdc' }}>
             <div style={{ flex: '1 0 0' }}>
+              <DialogReadOnly defaultValue="hi" label="Dance Entry" />
               <DialogReadOnly defaultValue="hi" label="Dance Title" />
               <DialogReadOnly defaultValue="hi" label="Performers" />
               <DialogReadOnly defaultValue="hi" label="Choreographer" />
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -86,9 +86,9 @@ class AdjudicationDialog extends React.Component {
           padding: '0',
           margin: '0'
         }}>
-          <Score type="subtotal" score={String(artisticMark)} scoreName="Artistic" />
-          <Score type="subtotal" score={String(technicalMark)} scoreName="Technical" />
-          <Score type="total" score={String(cumulativeMark)} scoreName="Score" />
+          <Score type="subtotal" score={artisticMark} scoreName="Artistic" />
+          <Score type="subtotal" score={technicalMark} scoreName="Technical" />
+          <Score type="total" score={cumulativeMark} scoreName="Score" />
         </DialogActions>
       </React.Fragment>
     );
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -122,7 +122,16 @@ class AdjudicationDialog extends React.Component {
 AdjudicationDialog.propTypes = {
   adjudicationId: PropTypes.string.isRequired,
   collectionName: PropTypes.string.isRequired,
-  currentValues: PropTypes.shape().isRequired
+  currentValues: PropTypes.shape({
+    artisticMark: PropTypes.number,
+    audio: PropTypes.bool,
+    choreoAward: PropTypes.number,
+    cumulativeMark: PropTypes.number,
+    notes: PropTypes.string,
+    judgeName: PropTypes.string,
+    specialAward: PropTypes.bool,
+    technicalMark: PropTypes.number
+  }).isRequired
 };
 
 export default withStyles(styles)(AdjudicationDialog);
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -32,17 +32,17 @@ class AdjudicationDialog extends React.Component {
 
   render() {
     const { adjudicationId, collectionName, currentValues } = this.props;
+    const { open, view } = this.state;
     const {
       artisticMark,
-      technicalMark,
+      audio,
+      choreoAward,
       cumulativeMark,
       judgeName,
       notes,
-      audio,
       specialAward,
-      choreoAward
+      technicalMark
     } = currentValues;
-    const { open, view } = this.state;
 
     const viewForm = (
       <React.Fragment>
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -73,13 +73,12 @@ class AdjudicationDialog extends React.Component {
               Special Award
             </div>
           )}
-          {choreoAward
-            && (
-              <div>
-                <LensIcon fontSize="inherit" color="inherit" style={{ color: 'purple' }} />
-                Choreography Award
-              </div>
-            )}
+          {choreoAward && (
+            <div>
+              <LensIcon fontSize="inherit" color="inherit" style={{ color: 'purple' }} />
+              Choreography Award
+            </div>
+          )}
         </div>
         <DialogActions style={{
           display: 'flex',
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -31,7 +31,7 @@ class AdjudicationDialog extends React.Component {
   }
 
   render() {
-    const { currentValues } = this.props;
+    const { adjudicationId, collectionName, currentValues } = this.props;
     const {
       artisticMark,
       technicalMark,
@@ -104,9 +104,13 @@ class AdjudicationDialog extends React.Component {
         handleView={this.handleView} />
     );
 
+    console.log(adjudicationId, collectionName)
+
     return (
       <DFDialog open={open} formType="edit" buttonTitle="edit" onClick={this.handleClickOpen} onClose={this.handleClose}>
         <DialogHeader
+          collectionName={collectionName}
+          docId={adjudicationId}
           edit={view}
           onEditClick={this.handleView}
           shouldShowDropdown
@@ -118,6 +122,8 @@ class AdjudicationDialog extends React.Component {
 }
 
 AdjudicationDialog.propTypes = {
+  adjudicationId: PropTypes.string.isRequired,
+  collectionName: PropTypes.string.isRequired,
   currentValues: PropTypes.shape().isRequired
 };
 
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -49,7 +49,6 @@ class AdjudicationDialog extends React.Component {
         <div style={{ display: 'flex', flexFlow: 'column', margin: '35px' }}>
           <div style={{ display: 'flex', justifyContent: 'space-around', borderBottom: '1px solid #dcdcdc' }}>
             <div style={{ flex: '1 0 0' }}>
-              <DialogReadOnly defaultValue="hi" label="Dance Entry" />
               <DialogReadOnly defaultValue="hi" label="Dance Title" />
               <DialogReadOnly defaultValue="hi" label="Performers" />
               <DialogReadOnly defaultValue="hi" label="Choreographer" />
@@ -68,17 +67,16 @@ class AdjudicationDialog extends React.Component {
           <DialogReadOnly fullWidth label="Notes" defaultValue={notes} />
           {audio && <AudioPlay fileName="file_test.mp3" time="3:07" />}
           <FormHelperText>Award Considerations</FormHelperText>
-          {specialAward
-            && (
-              <div>
-                <LensIcon fontSize="inherit" color="inherit" style={{ color: 'purple' }} />
-                Special Award
-              </div>
-            )}
+          {specialAward && (
+            <div>
+              <LensIcon fontSize="inherit" color="primary" />
+              Special Award
+            </div>
+          )}
           {choreoAward
             && (
               <div>
-                <LensIcon fontSize="inherit" color="primary" />
+                <LensIcon fontSize="inherit" color="inherit" style={{ color: 'purple' }} />
                 Choreography Award
               </div>
             )}
@@ -89,19 +87,21 @@ class AdjudicationDialog extends React.Component {
           padding: '0',
           margin: '0'
         }}>
-          <Score type="subtotal" score={artisticMark} scoreName="Artistic" />
-          <Score type="subtotal" score={technicalMark} scoreName="Technical" />
-          <Score type="total" score={cumulativeMark} scoreName="Score" />
+          <Score type="subtotal" score={String(artisticMark)} scoreName="Artistic" />
+          <Score type="subtotal" score={String(technicalMark)} scoreName="Technical" />
+          <Score type="total" score={String(cumulativeMark)} scoreName="Score" />
         </DialogActions>
       </React.Fragment>
     );
 
     const editForm = (
       <AdjudicationForm
-        view={view}
+        adjudicationId={adjudicationId}
+        collectionName={collectionName}
         currentValues={currentValues}
+        handleView={this.handleView}
         onModalClose={this.handleClose}
-        handleView={this.handleView} />
+        view={view} />
     );
 
     return (
@@ -112,7 +112,7 @@ class AdjudicationDialog extends React.Component {
           edit={view}
           onEditClick={this.handleView}
           shouldShowDropdown
-          title={judgeName} />
+          title={judgeName || ''} />
         {view ? (viewForm) : (editForm)}
       </DFDialog>
     );
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -104,8 +104,6 @@ class AdjudicationDialog extends React.Component {
         handleView={this.handleView} />
     );
 
-    console.log(adjudicationId, collectionName)
-
     return (
       <DFDialog open={open} formType="edit" buttonTitle="edit" onClick={this.handleClickOpen} onClose={this.handleClose}>
         <DialogHeader
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -45,15 +45,19 @@ class AdjudicationForm extends React.Component {
 
   // TODO: handle submmission of the form
   handleSubmit = async () => {
+    const { adjudicationId, collectionName } = this.props;
     const { artisticMark, technicalMark } = this.state;
-    const cumulativeMark = artisticMark + technicalMark;
-    const collectionName = 'adjudications';
-    const adjudicationId = 1;
+    const cumulativeMark = (Number(artisticMark) + Number(technicalMark)) / 2;
+    const data = {
+      cumulativeMark,
+      ...this.state
+    };
     await updateData(
       collectionName,
       adjudicationId,
-      { artisticMark, technicalMark, cumulativeMark }
+      data
     );
+    this.handleModalClose();
   }
 
   handleModalClose = () => {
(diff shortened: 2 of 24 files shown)


src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -68,17 +67,16 @@ class AdjudicationDialog extends React.Component {
...
-                Special Award
-              </div>
-            )}
+          {specialAward && (
+            <div>
+              <LensIcon fontSize="inherit" color="primary" />
+              Special Award
+            </div>
+          )}
           {choreoAward
```
YangStephen: Like the change above, let's bring this && up a line to be with the `choreoAward`

src/components/events/EventForm.jsx
```diff
@@ -16,16 +18,16 @@ class EventForm extends React.Component {
     const { currentValues } = props;
 
     this.state = {
-      eventTitle: currentValues.eventTitle || '',
       eventDate: currentValues.eventDate || '',
-      numJudges: currentValues.numJudges || '',
-      disabled: true
+      eventTitle: currentValues.eventTitle || '',
+      disabled: true,
```
YangStephen: Nitty, let's bring disabled to the top for alphabetical order

src/components/events/EventsSection.jsx
```diff
@@ -48,18 +48,26 @@ class EventsSection extends React.Component {
   render() {
     const { events } = this.state;
     const headings = ['Event Title', 'Event Date', 'No. Dancers', 'No. Performances', 'No. Judges'];
+    const showEvents = Array.isArray(events) && events.length > 0;
```
YangStephen: You do not need the > 0 check.

src/components/events/EventDialog.jsx
```diff
@@ -46,18 +46,25 @@ class EventDialog extends React.Component {
...
+          formType={formType}
+          onModalClose={this.handleClose} />
       </DFDialog>
     );
   }
 }
 
 EventDialog.propTypes = {
-  currentValues: PropTypes.shape().isRequired,
+  currentValues: PropTypes.shape(),
```
YangStephen: If we're doing this .shape() function, this may be a point where we can start bringing in stronger assurances on our PropTypes. 

Ref: https://reactjs.org/docs/typechecking-with-proptypes.html, take a look at `optionalObjectWithShape `. For these shapes, should we add in the actual structure of currentValues if we have them? Seems like we already know what we're passing down.

src/components/events/EventForm.jsx
```diff
@@ -41,15 +43,19 @@ class EventForm extends React.Component {
...
-
-
+    const data = {
+      eventDate: new Date(eventDate),
+      eventTitle,
+      numJudges
+    };
     if (formType === 'new') {
-      await addData(collectionName, data, this.handleModalClose);
+      await addData(collectionName, data);
```
YangStephen: How are we doing the modal closing now? Does it not close by default?

src/components/events/EventForm.jsx
```diff
@@ -90,14 +96,14 @@ class EventForm extends React.Component {
 }
 
 EventForm.propTypes = {
-  classes: PropTypes.string.isRequired,
+  classes: PropTypes.shape().isRequired,
```
YangStephen: if this is a shape, let's strong assure on the structure if we can.

src/components/events/EventTableRow.jsx
```diff
@@ -51,22 +51,13 @@ class EventTableRow extends React.Component {
 }
 
 EventTableRow.propTypes = {
-  id: PropTypes.string,
-  eventTitle: PropTypes.string,
-  eventDate: PropTypes.string,
-  numDancers: PropTypes.string,
-  numPerformances: PropTypes.string,
-  numJudges: PropTypes.string
+  currentValues: PropTypes.shape(),
```
YangStephen: strong assurance on shape()

src/components/events/EventsSection.jsx
```diff
@@ -22,16 +23,15 @@ class EventsSection extends React.Component {
     db.collection('events').onSnapshot((querySnapshot) => {
       const events = [];
       querySnapshot.forEach((doc) => {
-        const { eventTitle, numJudges, date } = doc.data();
-        let eventDate;
-        if (date) {
-          eventDate = new Date(date.seconds * 1000).toLocaleDateString();
-        }
+        const { eventTitle, numJudges, eventDate } = doc.data();
+        const date = typeof eventDate === 'object'
```
YangStephen: `typeof evenDate === 'object'` seems dangerous: https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript

Depending on what exactly the check for obj is here. We can use lodash.isObject: https://stackoverflow.com/questions/34111902/why-do-lodashs-isobject-isplainobject-behave-differently-than-typeof-x

src/components/interface/EmptyStates.jsx
```diff
@@ -21,17 +21,17 @@ const EmptyState = ({ type, title, subtitle }) => {
...
+      <img src={image} alt="" />
       <p>{title}</p>
       <p>{subtitle}</p>
     </div>
   );
 };
 
 EmptyState.propTypes = {
+  subtitle: PropTypes.string.isRequired,
   type: PropTypes.oneOf(['event', 'adjudication', 'performance']).isRequired,
```
YangStephen: swap type and title please!

src/components/interface/SectionHeader.jsx
```diff
@@ -29,7 +28,7 @@ const SectionHeader = ({
 );
 
 SectionHeader.propTypes = {
-  classes: PropTypes.string.isRequired,
+  classes: PropTypes.shape().isRequired,
```
YangStephen: assurance on shape maybe?

src/components/interface/TableHeader.jsx
```diff
@@ -1,32 +1,13 @@
...
 import TableRow from '@material-ui/core/TableRow';
 import { withStyles } from '@material-ui/core/styles';
 import styles from '../styles';
-import Filter from './filter/index';
 
-
-const TableHeader = ({ headings, classes }) => (
+const TableHeader = ({ headings }) => (
   <TableHead>
-    <div style={{ display: 'flex', margin: '5px' }}>
```
YangStephen: Just to confirm, this is the DOM object you are moving out into a different class correct?

src/components/interface/TableSettings.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import SearchIcon from '@material-ui/icons/Search';
```
YangStephen: To maintain style, if we're segmenting all these imports, let's space them out.

src/components/performances/PerformancesSection.jsx
```diff
@@ -37,29 +39,33 @@ class PerformancesSection extends React.Component {
...
     return (
       <React.Fragment>
         <SectionHeader eventId={eventId} title="performance" />
+        <TableSettings />
         <Table>
           <TableHeader headings={headings} />
           <TableBody>
-            {(Array.isArray(performances) && performances.length)
-              && performances.map(performance => (
+            {showPerformances && performances.map((performance) => {
```
YangStephen: I'm concerned about this logic actually, if we do showPerformances = true but performances.length = 0, then nothing shows up here.

src/components/performances/PerformancesSection.jsx
```diff
@@ -37,29 +39,33 @@ class PerformancesSection extends React.Component {
...
+              const currentValues = pick(performance, keys);
+              return (
                 <PerformanceTableRow
-                  key={performance.id}
-                  {...performance}
+                  currentValues={currentValues}
                   eventId={eventId}
-                  collectionName={collectionName} />
-              ))}
+                  id={performance.id}
```
YangStephen: Because performance.id is used more than once, I'd almost recommend we just destructure the performance object for this id and only pass in id. Though i'd lean either way, what are your thoughts @livictoria

src/components/performances/PerformancesSection.jsx
```diff
@@ -37,29 +39,33 @@ class PerformancesSection extends React.Component {
...
+                  key={performance.id} />
+              );
+            })}
           </TableBody>
         </Table>
         {
-          (!performances) && (
+          (!showPerformances) && (
```
YangStephen: We may want to also include this "empty performances page" if the array size is 0?

src/components/performances/PerformanceTableRow.jsx
```diff
@@ -12,77 +12,44 @@ class PerformanceTableRow extends React.Component {
...
-      danceStyle,
-      performers,
-      choreographers,
-      groupSize,
+      currentValues,
       eventId
     } = this.props;
-    const currentValues = {
+    const {
+      danceEntry,
```
YangStephen: not sure about this naming, it used to be an id, now its an entry. Is this the Dance Entry value (0, 1, ...)?

src/components/performances/PerformanceForm.jsx
```diff
@@ -115,15 +110,14 @@ class PerformanceForm extends React.Component {
 }
 
 PerformanceForm.propTypes = {
-  classes: PropTypes.string.isRequired,
+  classes: PropTypes.shape().isRequired,
```
YangStephen: Since this is required, lets see if we can tightly assert on it's structure

src/components/performances/PerformanceForm.jsx
```diff
@@ -17,7 +18,7 @@ class PerformanceForm extends React.Component {
     const { currentValues } = props;
 
     this.state = {
-      danceEntry: currentValues.danceEntry || 0,
+      danceEntry: currentValues.danceEntry || '',
```
YangStephen: Hm, if this danceEntry is actually the integer value we input on creation, shouldnt we keep it's default state an integer value, or is this just to hide the value if it doesn't exist?

src/components/interface/dialog/DialogHeader.jsx
```diff
@@ -29,7 +29,7 @@ const DialogHeader = ({
 export default withStyles(styles)(DialogHeader);
 
 DialogHeader.propTypes = {
-  classes: PropTypes.string.isRequired,
+  classes: PropTypes.shape().isRequired,
```
YangStephen: assert on structure

src/components/events/EventsSection.jsx
```diff
@@ -48,18 +48,26 @@ class EventsSection extends React.Component {
...
+                <EventTableRow
+                  currentValues={currentValues}
+                  id={event.id}
+                  key={event.id} />);
+            })
+          }
           </TableBody>
         </Table>
-        {!events && (
+        {!showEvents && (
```
YangStephen: Shouldn't it be `!showEvents || events.length == 0`? Otherwise, we just get an empty page :(

src/components/adjudications/AdjudicationTableRow.jsx
```diff
@@ -6,61 +6,50 @@ import LensIcon from '@material-ui/icons/Lens';
...
+        <p>
+          {audio ? 'yes' : 'no'}
+        </p>
+      </TableCell>
+      <TableCell>
+        {cumulativeMark}
+      </TableCell>
+      <TableCell>
+        {specialAward && <LensIcon fontSize="inherit" color="primary" />}
+        {choreoAward && <LensIcon fontSize="inherit" style={{ color: 'purple' }} />}
```
YangStephen: We just swapped the colors here, was that intended? `specialAward` used to have the color purple styling, and now it has the `primary` color tag.

src/components/adjudications/AdjudicationsSection.jsx
```diff
@@ -42,27 +42,27 @@ class AdjudicationsSection extends React.Component {
...
+                <AdjudicationTableRow
+                  collectionName={collectionName}
+                  currentValues={currentValues}
+                  key={rowProps.id}
+                  id={rowProps.id} />);
+            })}
           </TableBody>
         </Table>
-        {!adjudications && (
+        {!showAdjudications && (
```
YangStephen: Note: Let's do a check on adjudications.length

src/components/events/EventForm.jsx
```diff
@@ -41,9 +43,13 @@ class EventForm extends React.Component {
 
   handleSubmit = async () => {
     const { eventId, formType } = this.props;
+    const { eventTitle, eventDate, numJudges } = this.state;
     const collectionName = 'events';
-    const data = _.omit(this.state, 'disabled');
-
+    const data = {
+      eventDate: new Date(eventDate),
```
YangStephen: What is this behaviour if the eventDate is `''`? That's our default above, I'm not sure if this is safe behaviour for us to do. It appears to still be the same object as `_.omit(this.state, 'disabled');`.

I'm curious if maybe we could just keep this line as is and do the `new Date` somewhere else, maybe even just store the state as a `Date` object

src/components/adjudications/AdjudicationTableRow.jsx
livictoria: @YangStephen yep that was intended, I realized I swapped the colors up, from the zeplin designs `specialAward` is red (`primary` color tag) and `choreoAward` is purple

src/components/adjudications/AdjudicationsSection.jsx
livictoria: @YangStephen not sure if we're on the same page, but `showAdjudications` comes from line 45:

`const showAdjudications = Array.isArray(adjudications) && adjudications.length > 0;` 

which checks the adjudications.length, is this what you mean?

src/components/performances/PerformanceTableRow.jsx
livictoria: @YangStephen `danceEntry` is the entry id (0,1,2,...) for a specific performance, where as the `id` is the document id for the performance (a unique identifier we use for the sepcific docemnt in the performances collection in firestore i.e. `WH9oJH8PVT2coTECdTSG`), both values are still there, we just pass the id as a prop and the danceEntry inside the `currentValues` object

src/components/interface/SectionHeader.jsx
livictoria: @YangStephen  would like to sync with you on this, not sure how to add assurance on the classes object

src/components/events/EventForm.jsx
YangStephen: Resolved for now, backlogging error handling.

src/components/interface/TableSettings.jsx
```diff
@@ -0,0 +1,32 @@
...
+import Input from '@material-ui/core/Input';
+import { withStyles } from '@material-ui/core/styles';
+import styles from '../styles';
+import Filter from './filter/index';
+
+// TODO: implement filter components for award considerations
+const TableSettings = ({ classes }) => (
```
YangStephen: TableFilters

src/components/performances/PerformanceForm.jsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import omit from 'lodash/omit';
+
 import { withStyles } from '@material-ui/core/styles';
 import DialogActions from '@material-ui/core/DialogActions';
-import _ from 'lodash';
-import updateData from '../../firebase/utils/updateData';
 
+import updateData from '../../firebase/utils/updateData';
```
YangStephen: move this below add data please

src/components/adjudications/AdjudicationForm.jsx
```diff
@@ -45,15 +45,19 @@ class AdjudicationForm extends React.Component {
 
   // TODO: handle submmission of the form
   handleSubmit = async () => {
+    const { adjudicationId, collectionName } = this.props;
     const { artisticMark, technicalMark } = this.state;
-    const cumulativeMark = artisticMark + technicalMark;
-    const collectionName = 'adjudications';
-    const adjudicationId = 1;
+    const cumulativeMark = (Number(artisticMark) + Number(technicalMark)) / 2;
```
YangStephen: I'd like us to verify if this is a string or a Number, as we are casting both.

src/components/events/EventForm.jsx
```diff
@@ -41,15 +43,19 @@ class EventForm extends React.Component {
 
   handleSubmit = async () => {
     const { eventId, formType } = this.props;
+    const { eventTitle, eventDate, numJudges } = this.state;
```
YangStephen: Switch eventDate & eventTitle

src/components/events/EventForm.jsx
YangStephen: Let's do this for now because there is a default behaviour until we integrate in a date picker

src/components/events/EventsSection.jsx
```diff
@@ -48,18 +49,26 @@ class EventsSection extends React.Component {
...
-            {(Array.isArray(events) && events.length)
-              && events.map(event => (<EventTableRow key={event.id} {...event} />))
-            }
+            {showEvents && events.map((event) => {
+              const keys = ['eventTitle', 'eventDate', 'numJudges'];
+              const currentValues = pick(event, keys);
+              return (
+                <EventTableRow
+                  currentValues={currentValues}
+                  id={event.id}
```
YangStephen: Let's destructure this

src/components/interface/EmptyStates.jsx
```diff
@@ -21,17 +21,17 @@ const EmptyState = ({ type, title, subtitle }) => {
...
     </div>
   );
 };
 
 EmptyState.propTypes = {
-  type: PropTypes.oneOf(['event', 'adjudication', 'performance']).isRequired,
-  title: PropTypes.string.isRequired,
-  subtitle: PropTypes.string.isRequired
+  subtitle: PropTypes.string.isRequired,
+  type: PropTypes.oneOf(['adjudication', 'event', 'performance']).isRequired,
```
YangStephen: Switch title and type

src/components/interface/dialog/Score.jsx
```diff
@@ -34,6 +34,6 @@ export default Score;
 
 Score.propTypes = {
   type: PropTypes.oneOf(['subtotal', 'total']).isRequired,
-  score: PropTypes.number.isRequired,
+  score: PropTypes.string.isRequired,
```
YangStephen: switch this back to number

src/components/performances/PerformanceForm.jsx
```diff
@@ -53,14 +48,14 @@ class PerformanceForm extends React.Component {
   }
 
   handleSubmit = async () => {
-    const { performanceId, formType, eventId } = this.props;
-    const collectionName = `events/${eventId}/performances`;
-    const data = _.omit(this.state, 'disabled');
+    const { collectionName, performanceId, formType } = this.props;
```
YangStephen: swap performanceId and formType

src/components/performances/PerformanceForm.jsx
```diff
@@ -115,15 +110,23 @@ class PerformanceForm extends React.Component {
...
 PerformanceForm.propTypes = {
-  classes: PropTypes.string.isRequired,
-  currentValues: PropTypes.shape(),
-  eventId: PropTypes.string.isRequired,
+  classes: PropTypes.shape().isRequired,
+  currentValues: PropTypes.shape({
+    academicLevel: PropTypes.string,
+    choreographers: PropTypes.string,
+    danceEntry: PropTypes.number,
+    danceTitle: PropTypes.string,
```
YangStephen: danceTitle and danceStyle switch

src/components/performances/PerformanceTableRow.jsx
```diff
@@ -12,77 +12,57 @@ class PerformanceTableRow extends React.Component {
...
 
 PerformanceTableRow.propTypes = {
-  id: PropTypes.string,
-  collectionName: PropTypes.string.isRequired,
-  entryId: PropTypes.string,
+  currentValues: PropTypes.shape({
+    academicLevel: PropTypes.string,
+    choreographers: PropTypes.string,
+    danceEntry: PropTypes.number,
+    danceTitle: PropTypes.string,
```
YangStephen: switch here too

src/components/performances/PerformancesSection.jsx
```diff
@@ -37,29 +39,34 @@ class PerformancesSection extends React.Component {
   }
 
   render() {
-    const headings = ['Dance Title', 'Dance Entry', 'School', 'Acaademic Level', 'Level of Competition', 'Dance Style', 'Dance Size'];
     const { performances } = this.state;
     const { match: { params: { eventId }}} = this.props;
-    const collectionName = `events/${eventId}/performances`;
+    const headings = ['Dance Title', 'Dance Entry', 'School', 'Acaademic Level', 'Level of Competition', 'Dance Style', 'Dance Size'];
```
YangStephen: "Academic Level"

src/components/adjudications/AdjudicationsSection.jsx
```diff
@@ -36,21 +38,31 @@ class AdjudicationsSection extends React.Component {
   }
 
   render() {
-    const headings = ['Judge', 'Audio', 'Cummulative Score', 'Awards'];
+    const { match: { params: { eventId, performanceId }}} = this.props;
     const { adjudications } = this.state;
+    const collectionName = `events/${eventId}/performances/${performanceId}/adjudications`;
+    const headings = ['Judge', 'Audio', 'Cummulative Score', 'Awards'];
```
YangStephen: Note here

PR #26, opened 2018-12-30
Display setting options in performance dropdown

livictoria:
* maps through the different settings choices as drop-down options in performance from.
* FYI: currently blocked by #25


src/components/settings/SettingData.jsx
```diff
@@ -10,23 +10,31 @@ class SettingData extends React.Component {
...
-        {`${optionName} `}
-        <CloseIcon style={{ cursor: 'pointer' }} onClick={this.handleAlert} />
-        {alert && (<AlertDialog optionName={optionName} />)}
+      <div style={{ width: '100%', marginTop: '5px', marginBottom: '5px' }}>
+        <div style={{ clear: 'both', float: 'left', width: 'auto' }}>
+          {`${optionName} `}
+        </div>
+        <div style={{ float: 'right', width: 'auto' }}>
+          <CancelIcon fontSize="inherit" style={{ cursor: 'pointer', color: '#99999', verticalAlign: 'middle' }} onClick={this.handleShowAlert} />
+          <AlertDialog showAlert={showAlert} {...this.props} onClose={this.handleCloseAlert} />
```
YangStephen: Can we just pass optionName explicitly rather than destructure an entire props object should the object change in the future.

PR #29, opened 2019-01-03
Fixes Section Header Error

livictoria:
Passes `performanceDialog` or `eventDialog` as a prop into section header to fix undefined `eventId` error.

### DIFF
--- a/src/components/interface/SectionHeader.jsx
+++ b/src/components/interface/SectionHeader.jsx
@@ -5,38 +5,32 @@ import { withStyles } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
 
 import styles from '../styles';
-import EventDialog from '../events/EventDialog';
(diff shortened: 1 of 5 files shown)


src/components/interface/SectionHeader.jsx
```diff
@@ -5,38 +5,32 @@ import { withStyles } from '@material-ui/core/styles';
...
     <Typography variant="h3">
       {`${title}s`}
     </Typography>
-    <div className={classes.sectionHeaderAction}>
-      {showNew && (title === 'event'
-        ? (<EventDialog formType="new" />)
-        : (<PerformanceDialog eventId={eventId} formType="new" />)
-      )}
-    </div>
+    {!!renderNewButton && (
```
YangStephen: Let's check if this `!!renderNewButton` is required, otherwise it looks good!

PR #30, opened 2019-01-03
Filter-Search: Adding in the filter function

YangStephen:
The filters are currently hardcoded but this is a guide as to how the callback should be triggered. The function call should receive a `filters` input

cc: @livictoria


livictoria:
This can be closed now that the implementation is integrated into #32

PR #31, opened 2019-01-03
Dropdown for settings / logout, back button & event deletion

YangStephen:
Introduced SettingsDropdown which gives you 2 options (Settings / Logout) in the top left corner.
Introduced a UI back button to improve navigation UX.
Events can now be deleted
screen shot 2019-01-03 at 12 21 43 am
screen shot 2019-01-03 at 12 19 52 am

### DIFF
--- a/src/components/interface/SectionHeader.jsx
+++ b/src/components/interface/SectionHeader.jsx
@@ -6,39 +6,39 @@ import { withStyles } from '@material-ui/core/styles';
 import ArrowLeft from '@material-ui/icons/ArrowLeft';
 import Typography from '@material-ui/core/Typography';
 
+import Button from './Button';
 import styles from '../styles';
 
 const SectionHeader = ({
   classes,
+  history,
   renderNewButton,
   title
-}) => (
-  <div className={classes.sectionHeaderWrapper}>
-    <div className={classes.sectionHeaderBackButton}>
-      <Button type="default" onClick={() => history.goBack()}>
-        <ArrowLeft />
-      </Button>
+}) => {
+  const goBack = () => { history.goBack(); };
+  return (
+    <div className={classes.sectionHeaderWrapper}>
+      <div className={classes.sectionHeaderBackButton}>
+        <Button type="default" onClick={goBack}>
+          <ArrowLeft />
+        </Button>
+      </div>
+      <Typography variant="h3">
+        {`${title}s`}
+      </Typography>
+      {!!renderNewButton && (
+        <div className={classes.sectionHeaderAction}>
+          {renderNewButton}
+        </div>)
+      }
     </div>
-    <Typography variant="h3">
-      {`${title}s`}
-    </Typography>
-    {!!renderNewButton && (
-      <div className={classes.sectionHeaderAction}>
-        {renderNewButton}
-      </div>)
-    }
-  </div>
-);
+  );
+};
 
 SectionHeader.propTypes = {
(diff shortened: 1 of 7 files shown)


src/components/interface/settings/SettingsDropdown.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+  handleClose = () => {
+    this.setState({ anchorEl: undefined });
+  };
+
+  handleLogout = () => {
+    auth.signOut();
+  };
+
+  handleSettingsRedirect = (event) => {
```
livictoria: it seems as if we don't utilize this `handleSettingsRedirect` function, we can remove this right?

src/components/interface/settings/SettingsDropdown.jsx
```diff
@@ -0,0 +1,72 @@
...
+    const { history } = this.props;
+    history.push('/settings');
+  };
+
+  render() {
+    const { anchorEl } = this.state;
+    return (
+      <div style={{
```
livictoria: can we move this styling into the styles file

src/components/interface/settings/SettingsDropdown.jsx
YangStephen: Ah so this was a leftover function after a refactor. The dropdown here has 2 buttons: The settings button and the Logout function. The logout function doesn't require a redirect, it just needs to invalidate the `auth` object.

Any thoughts on if we should keep using `<Link to="/settings">` or use a handler function `onClick={handleSettingsRedirect}`?

src/components/interface/settings/SettingsDropdown.jsx
```diff
@@ -0,0 +1,63 @@
...
+          aria-haspopup="true"
+          onClick={this.handleClick}
+          style={{ cursor: 'pointer' }} />
+        <Menu
+          id="simple-menu"
+          anchorEl={anchorEl}
+          open={Boolean(anchorEl)}
+          onClose={this.handleClose}>
+          <Link to="/settings">
+            <MenuItem onClick={this.handleClose}>Settings</MenuItem>
```
livictoria: For the dropdown, I think Settings is wrapped with a `link` while Logout isn't causing it to have different styling: i.e. 
image

if you add 
```
a:focus {
  outline: none;
}
```
to the `index.css` it should fix the styling inconsistency

src/components/interface/SectionHeader.jsx
```diff
@@ -1,30 +1,43 @@
...
-      </div>)
-    }
-  </div>
-);
+}) => {
+  const goBack = () => { history.goBack(); };
+  return (
+    <div className={classes.sectionHeaderWrapper}>
+      <div className={classes.sectionHeaderBackButton}>
+        <Button type="default" onClick={goBack}>
```
livictoria: We should also add a back button to the settings page. If we extract the components that make up the back button here and put it into its own component i.e. `BackButton`. then we can use here and it in the Settings Page

livictoria: @YangStephen actually I just realized that Settings is missing a Section Header LOOOL so it's fine we can actually just leave that there and actually just add a section header to settings i.e.
image

And then the extraction of the back button is a nice to have but not needed

src/components/interface/settings/SettingsDropdown.jsx
YangStephen: This is resolved in the latest push.

PR #47, opened 2019-01-24
Add Search Functionality

livictoria:
**What this PR does:** Adds search functionality to search for performances
Fixes Issue #42

### DIFF
--- a/src/components/interface/filter/index.jsx
+++ b/src/components/interface/filter/index.jsx
@@ -25,7 +25,8 @@ class Filter extends React.Component {
         danceSize: [],
         danceStyle: [],
         school: []
-      }
+      },
+      search: undefined
     };
   }
 
@@ -40,11 +41,11 @@ class Filter extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const { filtered: prevFiltered } = prevState;
-    const { filtered: currFiltered } = this.state;
+    const { filtered: prevFiltered, search: prevSearch } = prevState;
+    const { filtered: currFiltered, search: currSearch } = this.state;
     const { handleFilters } = this.props;
-    if (!isEqual(prevFiltered, currFiltered)) {
-      handleFilters(currFiltered);
+    if (!isEqual(prevFiltered, currFiltered) || prevSearch !== currSearch) {
+      handleFilters(currFiltered, currSearch);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/interface/filter/index.jsx
```diff
@@ -40,11 +41,11 @@ class Filter extends React.Component {
...
 
   componentDidUpdate(prevProps, prevState) {
-    const { filtered: prevFiltered } = prevState;
-    const { filtered: currFiltered } = this.state;
+    const { filtered: prevFiltered, search: prevSearch } = prevState;
+    const { filtered: currFiltered, search: currSearch } = this.state;
     const { handleFilters } = this.props;
-    if (!isEqual(prevFiltered, currFiltered)) {
-      handleFilters(currFiltered);
+    if (!isEqual(prevFiltered, currFiltered) || prevSearch !== currSearch) {
```
YangStephen: *apparantly* you could actually use !isEqual() for prevSearch and currSearch, it looks like it covers comparing between String objects.

src/components/performances/PerformancesSection.jsx
```diff
@@ -44,13 +44,22 @@ class PerformancesSection extends React.Component {
...
     const { performances } = this.state;
     const keys = this.getFilterKeys(filters);
-    const filteredPerformances = performances.filter(
-      performance => keys.every(key => filters[key].includes(performance[key]))
-    );
+    const filterFunction = (performance) => {
+      const danceTitle = performance.danceTitle.toLowerCase();
+      const passFilter = keys.every(key => filters[key].includes(performance[key]));
+
+      if (search && search.length > 0) {
```
YangStephen: Could we rename some of these variables to be more descriptive. I'm not too sure what "passFilter" is and "search" is kind of generic.

Also, it would be really nice to have some comments especially here. This block of code is a bit less intuitive.

src/components/interface/filter/index.jsx
```diff
@@ -25,7 +25,8 @@ class Filter extends React.Component {
         danceSize: [],
         danceStyle: [],
         school: []
-      }
+      },
```
YangStephen: plus, one minor thing, let's alphabetize this state. Filtered above options.

YangStephen:
@livictoria Could address the above comments and get this merged in? :)

PR #49, opened 2019-01-26
Audio playback

meganniu:
Adds audio play functionality to the adjudication dialog box.
Fixed issue #43

### DIFF
--- a/cors.json
+++ b/cors.json
@@ -1,0 +1,7 @@
+[
+    {
+      "origin": ["*"],
+      "method": ["GET"],
+      "maxAgeSeconds": 3600
+    }
+  ]
--- a/cors.json
+++ b/cors.json
@@ -1,7 +1,0 @@
-[
-    {
-      "origin": ["*"],
-      "method": ["GET"],
-      "maxAgeSeconds": 3600
-    }
-  ]
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -69,7 +69,7 @@ class AdjudicationDialog extends React.Component {
           </div>
           <DialogReadOnly fullWidth label="Notes" defaultValue={notes} />
           {/*audio && <AudioPlay fileName="file_test.mp3" time="3:07" />*/}
-          {audio && <AudioPlay currentValues={currentValues}/>}
+          {audio && <AudioPlay audio={audio}/>}
           <FormHelperText>Award Considerations</FormHelperText>
           {specialAward && (
             <div>
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -13,8 +13,6 @@ import DialogHeader from '../interface/dialog/DialogHeader';
 import AdjudicationForm from './AdjudicationForm';
 import Score from '../interface/dialog/Score';
 
-
-
 class AdjudicationDialog extends React.Component {
   state = { open: false, view: true };
 
@@ -68,8 +66,7 @@ class AdjudicationDialog extends React.Component {
             </div>
           </div>
           <DialogReadOnly fullWidth label="Notes" defaultValue={notes} />
-          {/*audio && <AudioPlay fileName="file_test.mp3" time="3:07" />*/}
-          {audioURL !== "" && <AudioPlay audioURL={audioURL}/>}
+          {audioURL && <AudioPlay audioURL={audioURL}/>}
           <FormHelperText>Award Considerations</FormHelperText>
           {specialAward && (
             <div>
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -127,8 +127,8 @@ AdjudicationDialog.propTypes = {
   collectionName: PropTypes.string.isRequired,
   currentValues: PropTypes.shape({
     artisticMark: PropTypes.number,
-    audio: PropTypes.bool,
-    choreoAward: PropTypes.number,
+    audio: PropTypes.string,
+    choreoAward: PropTypes.bool,
     cumulativeMark: PropTypes.number,
     notes: PropTypes.string,
     judgeName: PropTypes.string,
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -37,7 +37,7 @@ class AdjudicationDialog extends React.Component {
     const { open, view } = this.state;
     const {
       artisticMark,
-      audio,
+      audioURL,
       choreoAward,
       cumulativeMark,
       judgeName,
@@ -69,7 +69,7 @@ class AdjudicationDialog extends React.Component {
           </div>
           <DialogReadOnly fullWidth label="Notes" defaultValue={notes} />
           {/*audio && <AudioPlay fileName="file_test.mp3" time="3:07" />*/}
-          {audio && <AudioPlay audio={audio}/>}
+          {audioURL !== "" && <AudioPlay audioURL={audioURL}/>}
           <FormHelperText>Award Considerations</FormHelperText>
           {specialAward && (
             <div>
@@ -127,7 +127,7 @@ AdjudicationDialog.propTypes = {
   collectionName: PropTypes.string.isRequired,
   currentValues: PropTypes.shape({
     artisticMark: PropTypes.number,
-    audio: PropTypes.string,
+    audioURL: PropTypes.string,
     choreoAward: PropTypes.bool,
     cumulativeMark: PropTypes.number,
     notes: PropTypes.string,
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -13,6 +13,8 @@ import DialogHeader from '../interface/dialog/DialogHeader';
 import AdjudicationForm from './AdjudicationForm';
 import Score from '../interface/dialog/Score';
 
+
+
 class AdjudicationDialog extends React.Component {
   state = { open: false, view: true };
 
@@ -66,7 +68,8 @@ class AdjudicationDialog extends React.Component {
             </div>
           </div>
           <DialogReadOnly fullWidth label="Notes" defaultValue={notes} />
-          {audio && <AudioPlay fileName="file_test.mp3" time="3:07" />}
+          {/*audio && <AudioPlay fileName="file_test.mp3" time="3:07" />*/}
+          {audio && <AudioPlay currentValues={currentValues}/>}
           <FormHelperText>Award Considerations</FormHelperText>
           {specialAward && (
             <div>
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -104,7 +104,7 @@ class AdjudicationForm extends React.Component {
           </div>
           <DialogInput fullWidth disabled multiline name="notes" value={notes} label="Notes" />
           <CheckBox label="Award Considerations" choices={choices} onChange={this.handleCheckedAward} />
-        </div>
+        </div>choices
         <div className={classes.dfdialog_footer}>
           <DialogActions>
             <Button type="default" onClick={this.handleCancel}>
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -104,7 +104,7 @@ class AdjudicationForm extends React.Component {
           </div>
           <DialogInput fullWidth disabled multiline name="notes" value={notes} label="Notes" />
           <CheckBox label="Award Considerations" choices={choices} onChange={this.handleCheckedAward} />
-        </div>choices
+        </div>
         <div className={classes.dfdialog_footer}>
           <DialogActions>
             <Button type="default" onClick={this.handleCancel}>
--- a/src/components/adjudications/AdjudicationTableRow.jsx
+++ b/src/components/adjudications/AdjudicationTableRow.jsx
@@ -26,7 +26,7 @@ const AdjudicationTableRow = ({
       </TableCell>
       <TableCell>
         <p>
-          {audioURL !== "" ? 'yes' : 'no'}
+          {audioURL ? 'yes' : 'no'}
         </p>
       </TableCell>
       <TableCell>
--- a/src/components/adjudications/AdjudicationTableRow.jsx
+++ b/src/components/adjudications/AdjudicationTableRow.jsx
@@ -12,7 +12,7 @@ const AdjudicationTableRow = ({
   id
 }) => {
   const {
-    audio,
+    audioURL,
     choreoAward,
     cumulativeMark,
     judgeName,
@@ -26,7 +26,7 @@ const AdjudicationTableRow = ({
       </TableCell>
       <TableCell>
         <p>
-          {audio ? 'yes' : 'no'}
+          {audioURL !== "" ? 'yes' : 'no'}
         </p>
       </TableCell>
       <TableCell>
@@ -49,7 +49,7 @@ const AdjudicationTableRow = ({
 AdjudicationTableRow.propTypes = {
   currentValues: PropTypes.shape({
     artisticMark: PropTypes.number,
-    audio: PropTypes.string,
+    audioURL: PropTypes.string,
     choreoAward: PropTypes.bool,
     cumulativeMark: PropTypes.number,
     notes: PropTypes.string,
--- a/src/components/adjudications/AdjudicationTableRow.jsx
+++ b/src/components/adjudications/AdjudicationTableRow.jsx
@@ -49,8 +49,8 @@ const AdjudicationTableRow = ({
 AdjudicationTableRow.propTypes = {
   currentValues: PropTypes.shape({
     artisticMark: PropTypes.number,
-    audio: PropTypes.bool,
-    choreoAward: PropTypes.number,
+    audio: PropTypes.string,
+    choreoAward: PropTypes.bool,
     cumulativeMark: PropTypes.number,
     notes: PropTypes.string,
     judgeName: PropTypes.string,
--- a/src/components/interface/AudioPlay.jsx
+++ b/src/components/interface/AudioPlay.jsx
@@ -12,17 +12,22 @@ class AudioPlayer extends React.Component {
     super(props);
 
     this.state = {
-      audioFile: undefined
+      audioFile: undefined,
+      metadataLoaded: false
     };
   }
 
   loadAudio = () => {
+    //const {audioFile} = this.state;
     this.pathRef.getDownloadURL().then(function (url) {
       var xhr = new XMLHttpRequest();
       xhr.responseType = 'blob';
       xhr.onload = async function (event) {
         var blob = xhr.response;
         this.setState({ audioFile: new Audio(url) });
+        this.state.audioFile.addEventListener('loadeddata', () => {
+          this.setState({ metadataLoaded: true});
+        })
       }.bind(this);
       xhr.open('GET', url);
       xhr.send();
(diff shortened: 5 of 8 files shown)


src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
+      audioFile.pause();
+    }
+  }
 
+  formattedAudioDuration = () => {
+    const { audioFile } = this.state;
 
-  // TODO: create method to convert time to XX:XX format
+    let secs = audioFile.duration;
+    let formattedTime = Math.floor(secs / 60) + ":" + Math.floor(secs % 60);
```
YangStephen: These can all be `const`s. General React styling is we use const unless we need to be able to alter the object.

If I recall correctly, you could even use a const and still manipulate the object. `const` binds the reference rather than the value.

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
+    return formattedTime;
+  }
 
   render() {
-    const { fileName, time } = this.props;
-    const { disabled } = this.state;
+    const { audioURL } = this.props;
+    const { audioFile, metadataLoaded } = this.state;
+
+    const fileName = firebase.storage().ref().child(audioURL).name;
```
YangStephen: We can keep this declaration a line above.

src/components/interface/AudioPlay.jsx
```diff
@@ -33,20 +103,27 @@ class AudioPlayer extends React.Component {
...
-        <div>
-          <PlayCircleOutlineIcon style={{ float: 'right', marginLeft: '25px' }} color="primary" />
+        <div style={{flex: 0}}>
+          {
+            audioFile == undefined 
+              ? <Button disabled={true} type='transparent' onClick={this.handleClick}>
+                <PlayCircleOutlineIcon style={{ float: 'right', marginLeft: '25px' }} color="primary" />
+              </Button>
+              : <Button type='transparent' onClick={this.handleClick}>
+                <PlayCircleOutlineIcon style={{ float: 'right', marginLeft: '25px' }} color="primary" />
```
YangStephen: It looks like these 2 may be consolidated into 1 statement, A lot of the code here repeats. Maybe we could use a single declaration but just use some conditions to outline which parameters should be set and which ones shouldn't inline? :)

src/components/adjudications/AdjudicationForm.jsx
```diff
@@ -104,7 +104,7 @@ class AdjudicationForm extends React.Component {
           </div>
           <DialogInput fullWidth disabled multiline name="notes" value={notes} label="Notes" />
           <CheckBox label="Award Considerations" choices={choices} onChange={this.handleCheckedAward} />
-        </div>
+        </div>choices
```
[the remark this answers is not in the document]
meganniu: That's so weird I shouldn't even be edition AdjudicationForm.jsx
I'll remove choices from line 107 and make sure everything still works.

cors.json
```diff
@@ -0,0 +1,7 @@
+[
```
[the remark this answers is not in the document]
meganniu: Deleted cors.json

src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -13,6 +13,8 @@ import DialogHeader from '../interface/dialog/DialogHeader';
 import AdjudicationForm from './AdjudicationForm';
 import Score from '../interface/dialog/Score';
 
+
```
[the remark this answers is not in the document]
meganniu: Deleted lines

src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -66,7 +68,8 @@ class AdjudicationDialog extends React.Component {
             </div>
           </div>
           <DialogReadOnly fullWidth label="Notes" defaultValue={notes} />
-          {audio && <AudioPlay fileName="file_test.mp3" time="3:07" />}
+          {/*audio && <AudioPlay fileName="file_test.mp3" time="3:07" />*/}
```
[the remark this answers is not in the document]
meganniu: Deleted line 71

src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -66,7 +68,8 @@ class AdjudicationDialog extends React.Component {
...
+          {audioURL !== "" && <AudioPlay audioURL={audioURL}/>}
```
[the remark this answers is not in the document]
meganniu: Got it, thanks.

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
+
+    this.pathRef = firebase.storage().ref().child(audioURL);
+    if (audioFile == undefined) {
+      this.loadAudio();
+    }
+  }
+
+  componentWillUnmount() {
+    const { audioFile } = this.state;
+    if (audioFile != undefined) {
```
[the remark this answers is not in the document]
meganniu: Edited 61

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
 
   render() {
-    const { fileName, time } = this.props;
-    const { disabled } = this.state;
+    const { audioURL } = this.props;
+    const { audioFile, metadataLoaded } = this.state;
+
+    const fileName = firebase.storage().ref().child(audioURL).name;
+
+    let time = 'X:XX';
```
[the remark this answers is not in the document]
meganniu: Simplified time assignment

src/components/interface/AudioPlay.jsx
meganniu: Got it, thanks

src/components/interface/AudioPlay.jsx
meganniu: Done

src/components/interface/AudioPlay.jsx
meganniu: Done

src/components/adjudications/AdjudicationTableRow.jsx
```diff
@@ -26,7 +26,7 @@ const AdjudicationTableRow = ({
       </TableCell>
       <TableCell>
         <p>
-          {audio ? 'yes' : 'no'}
+          {audioURL !== "" ? 'yes' : 'no'}
```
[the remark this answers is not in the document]
meganniu: Done

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
 import PlayCircleOutlineIcon from '@material-ui/icons/PlayCircleOutline';
 import styles from '../styles';
+import Button from '../interface/Button';
+import * as firebase from 'firebase';
+
```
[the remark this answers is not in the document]
meganniu: Done

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
     this.state = {
-      disabled: false
+      audioFile: undefined,
+      metadataLoaded: false
     };
   }
 
-  // TODO: handle audio playing file
+  loadAudio = () => {
+    //const {audioFile} = this.state;
```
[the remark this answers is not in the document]
meganniu: Done

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
+      console.log("ERROR. Firebase audio file failed to load.");
+    });;
+  }
+
+  loadMetadata = () => {
+    const { audioFile } = this.state;
+
+    audioFile.preload = "auto";
+    audioFile.load();
+    //this.setState({metadataLoaded: true});
```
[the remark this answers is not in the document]
meganniu: Yeah, the function is outdated. I'll remove it.

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
+  }
+
+  componentDidMount() {
+    const { audioURL } = this.props;
+    const { audioFile } = this.state;
+
+    console.log("HERE");
```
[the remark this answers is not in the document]
meganniu: Done

src/components/interface/AudioPlay.jsx
```diff
@@ -3,24 +3,94 @@ import PropTypes from 'prop-types';
...
+
+    this.pathRef = firebase.storage().ref().child(audioURL);
+    if (audioFile == undefined) {
```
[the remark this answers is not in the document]
meganniu: Done

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
```
YangStephen: This should be with the material imports.

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
+
 import PlayCircleOutlineIcon from '@material-ui/icons/PlayCircleOutline';
```
YangStephen: group this with the material import on line 3.

Line 1,2 together. New Line. Line 3 and 5 together. New Line, then lines 6-8 together.

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
     this.state = {
-      disabled: false
+      audioFile: undefined,
+      metadataLoaded: false
     };
   }
 
-  // TODO: handle audio playing file
+  loadAudio = () => {
+    this.pathRef.getDownloadURL().then(function (url) {
```
YangStephen: Javascript style is that we use arrow functions instead of `function (...) { ... }`.

`.then( (url) => { ... } )`

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
+      var xhr = new XMLHttpRequest();
+      xhr.responseType = 'blob';
+      xhr.onload = async function (event) {
```
YangStephen: same here, let's make this into an arrow function if possible.

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
-      disabled: false
+      audioFile: undefined,
+      metadataLoaded: false
     };
   }
 
-  // TODO: handle audio playing file
+  loadAudio = () => {
+    this.pathRef.getDownloadURL().then(function (url) {
+      var xhr = new XMLHttpRequest();
```
YangStephen: So it looks like instead of XMLHttpRequest it *may* be possible to use Fetch instead.

Here's a blog post about it: https://medium.com/beginners-guide-to-mobile-web-development/the-fetch-api-2c962591f5c

src/components/interface/AudioPlay.jsx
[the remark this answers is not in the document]
YangStephen: It doesn't appear this change propagated. Also, just as a note: In Javascript, it is preferred if we ever do use a comparison like `==` to use `===`. `===` also does a type check on the objects. == will check for equality after type conversions, === will not do the conversion, so it assures on the type.

Ref: https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
 
+    if (audioFile) {
+      audioFile.pause();
+    }
+  }
+
+  handleClick = () => {
+    const { audioFile } = this.state;
+
+    if (audioFile.paused) {
```
YangStephen: NIT: Negotiable style but typically to save line space: 

```
if (...) {
    ...
} else { <-- Else on the same line as the if terminator.
   ...
}
```

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
+      audioFile.pause();
+    }
+  }
+
+  formattedAudioDuration = () => {
+    const { audioFile } = this.state;
 
-  // TODO: create method to convert time to XX:XX format
+    const secs = audioFile.duration;
+    const formattedTime = Math.floor(secs / 60) + ":" + Math.floor(secs % 60);
```
YangStephen: Can we get an extra new line between const declarations and a return statement?

src/components/interface/AudioPlay.jsx
```diff
@@ -33,20 +89,23 @@ class AudioPlayer extends React.Component {
...
         </div>
-        <div style={{ alignSelf: 'center', flex: 2 }}>
+        <div style={{ alignSelf: 'center', flex: 0 }}>
           {time}
         </div>
-        <div>
-          <PlayCircleOutlineIcon style={{ float: 'right', marginLeft: '25px' }} color="primary" />
+        <div style={{ flex: 0 }}>
+          {
+            <Button disabled={audioFile == undefined ? true : false} type='transparent' onClick={this.handleClick}>
```
YangStephen: Just to check my logic here: isn't `disabled={audioFile == undefined ? true :  false}` the same as `disabled={audioFile == undefined}` ?

Also: preferred to use `===`

src/components/interface/AudioPlay.jsx
```diff
@@ -33,20 +89,23 @@ class AudioPlayer extends React.Component {
...
+              <PlayCircleOutlineIcon style={{ float: 'right', marginLeft: '25px' }} color="primary" />
```
YangStephen: while we're at this, let's alphabetize these props.

src/components/interface/AudioPlay.jsx
YangStephen: Also can we alphabetize these props?

src/components/styles.js
```diff
@@ -147,6 +147,10 @@ const styles = theme => ({
     cursor: 'no-drop',
     color: '#fff!important'
   },
+  button_transparent: {
```
YangStephen: If this style is to make a button match the background by being white, we may just want to adjust the opacity?

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,82 @@
...
+    return formattedTime;
+  }
 
   render() {
-    const { fileName, time } = this.props;
-    const { disabled } = this.state;
+    const { audioURL } = this.props;
+    const { audioFile, metadataLoaded } = this.state;
+    const fileName = firebase.storage().ref().child(audioURL).name;
+    let time = (audioFile && metadataLoaded) ? this.formattedAudioDuration() : 'X:XX';
```
YangStephen: this can be a `const`

src/components/interface/AudioPlay.jsx
```diff
@@ -33,20 +89,23 @@ class AudioPlayer extends React.Component {
...
           {fileName}
         </div>
-        <div style={{ alignSelf: 'center', flex: 2 }}>
+        <div style={{ alignSelf: 'center', flex: 0 }}>
           {time}
         </div>
-        <div>
-          <PlayCircleOutlineIcon style={{ float: 'right', marginLeft: '25px' }} color="primary" />
+        <div style={{ flex: 0 }}>
+          {
```
YangStephen: And a one last comment, why are these JSX objects wrapped in `{ }`?

Typically we use `{ }` evaluating JS inside of these React nodes.

src/components/styles.js
meganniu: When I adjusted the opacity of the button, the opacity of picture on the button was also affected. I removed the `backgroundColor` property instead.

src/components/interface/AudioPlay.jsx
meganniu: I think wrapping the JSX objects in `{ }` was for organisation. I have removed them and alphabetised the properties.

src/components/interface/AudioPlay.jsx
meganniu: > Just to check my logic here: isn't `disabled={audioFile == undefined ? true : false}` the same as `disabled={audioFile == undefined}` ?

Yup! It's changed now to `disabled={audioFile == undefined}`

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,78 @@
...
+      metadataLoaded: false
     };
   }
 
-  // TODO: handle audio playing file
+  loadAudio = () => {
+    this.pathRef.getDownloadURL().then((url) => {
+      fetch(url).then(() => {
+        this.setState({ audioFile: new Audio(url) });
+        this.state.audioFile.addEventListener('loadeddata', () => {
```
YangStephen: It's nice in JS when we don't need to do object references at such a deep level, like `this.state.audiofile....`.

Would it be the same in this case if we did:
```
const audioFile = new Audio(url);
audioFile.addEventListener ....
this.setState({ audioFile }) // <- this actually is equivalent to setting {audioFile: audioFile}
```

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,78 @@
...
 
     this.state = {
-      disabled: false
+      audioFile: undefined,
+      metadataLoaded: false
     };
   }
 
-  // TODO: handle audio playing file
+  loadAudio = () => {
```
YangStephen: I would typically argue that these should be alphabeticized but I think in React, the lifecycle components are more important to keep at the top.

How would you feel about the following styling:

Let's move the react lifecycle components under the constructor alphabetically, then these arrow functions below that alphabetically. Finally the Render at the end?

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,78 @@
...
+    this.pathRef.getDownloadURL().then((url) => {
+      fetch(url).then(() => {
+        this.setState({ audioFile: new Audio(url) });
+        this.state.audioFile.addEventListener('loadeddata', () => {
+          this.setState({ metadataLoaded: true });
+        })
+      })
+    }).catch(function () {
+      console.log("ERROR. Firebase audio file failed to load.");
+    });;
```
YangStephen: Why is there a double ;

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,78 @@
...
+  }
+
+  componentDidMount() {
+    const { audioURL } = this.props;
+    const { audioFile } = this.state;
+
+    this.pathRef = firebase.storage().ref().child(audioURL);
```
YangStephen: Should we be re-creating this pathRef on every mount? If audioFile already exists, doesn't that mean that the pathRef already existed? Maybe we should only initialize this pathRef when we don't have an audioFile?

src/components/interface/AudioPlay.jsx
meganniu: Would it be better to put the arrow functions at the top then all the lifecycle components below with `render` at the end? That way all the lifecycle components including `render` will be grouped together.

src/components/interface/AudioPlay.jsx
meganniu: Whoops, typo. Good catch :)

src/components/interface/AudioPlay.jsx
meganniu: `{ }`  were left over from an old comment. Thanks!

src/components/interface/AudioPlay.jsx
meganniu: > const audioFile = new Audio(url); audioFile.addEventListener .... this.setState({ audioFile }) // <- this actually is equivalent to setting {audioFile: audioFile}
also works. I'll use it instead.

src/components/interface/AudioPlay.jsx
meganniu: Thanks for pointing this out. Turns out I only used `pathRef` once to get the download URL. I've done away with `pathRef` completely.

src/components/interface/AudioPlay.jsx
```diff
@@ -1,26 +1,78 @@
...
+
+    return formattedTime;
+  }
 
   render() {
-    const { fileName, time } = this.props;
-    const { disabled } = this.state;
+    const { audioURL } = this.props;
+    const { audioFile, metadataLoaded } = this.state;
+    const fileName = firebase.storage().ref().child(audioURL).name;
```
meganniu: Inefficient to retrieve the file at `audioURL` every time the component renders. `fileName` has been made into a `state` property and it is assigned when the component mounts.

PR #51, opened 2019-01-26
Performance Form data validation

meganniu:
Save button is only enabled if all fields of a Performance Form are filled

### DIFF
--- a/src/components/performances/PerformanceForm.jsx
+++ b/src/components/performances/PerformanceForm.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import omit from 'lodash/omit';
+import pick from 'lodash/pick';
 
 import { withStyles } from '@material-ui/core/styles';
 import DialogActions from '@material-ui/core/DialogActions';
@@ -19,18 +20,26 @@ class PerformanceForm extends React.Component {
     const { currentValues } = props;
 
     this.state = {
+      disabled: true,
       academicLevel: currentValues.academicLevel || '',
       choreographers: currentValues.choreographers || '',
       competitionLevel: currentValues.competitionLevel || '',
       danceEntry: currentValues.danceEntry,
       danceStyle: currentValues.danceStyle || '',
       danceTitle: currentValues.danceTitle || '',
-      disabled: false,
       options: {},
       performers: currentValues.performers || '',
       school: currentValues.school || '',
       size: currentValues.size || 0
     };
+  }
+
+  // disable save button if not all input fields are filled
+  static getDerivedStateFromProps(props, state) {
+    const values = pick(state, 
+      ['academicLevel', 'choreographers', 'competitionLevel', 'danceEntry', 
+      'danceStyle', 'danceTitle', 'performers', 'school', 'size']);
+    return { disabled: !(Object.keys(values).every(value => !!state[value])) };
   }
 
   componentDidMount() {
@@ -79,6 +88,7 @@ class PerformanceForm extends React.Component {
   render() {
     const { classes, formType } = this.props;
     const {
+      disabled,
       academicLevel,
       choreographers,
       competitionLevel,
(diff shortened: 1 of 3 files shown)


src/components/performances/PerformanceForm.jsx
```diff
@@ -171,7 +181,7 @@ class PerformanceForm extends React.Component {
             <Button type="default" onClick={this.handleCancel}>
               {formType === 'edit' ? 'cancel' : 'discard'}
             </Button>
-            <Button disabled={false} type="primary" onClick={this.handleSubmit}>
+            <Button disabled={disabled} type="primary" onClick={this.handleSubmit}>
```
YangStephen: let's reorder these props

src/components/performances/PerformanceForm.jsx
```diff
@@ -79,6 +88,7 @@ class PerformanceForm extends React.Component {
   render() {
     const { classes, formType } = this.props;
     const {
+      disabled,
```
YangStephen: order this alphabetically

src/components/performances/PerformanceForm.jsx
```diff
@@ -19,20 +20,28 @@ class PerformanceForm extends React.Component {
     const { currentValues } = props;
 
     this.state = {
+      disabled: true,
```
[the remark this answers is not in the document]
YangStephen: Also, could we name this something more specific. `disabledSubmit` or something? This seems like the entire pane should be disabled.

src/components/performances/PerformanceForm.jsx
```diff
@@ -19,20 +20,28 @@ class PerformanceForm extends React.Component {
...
       performers: currentValues.performers || '',
       school: currentValues.school || '',
       size: currentValues.size || 0
     };
   }
 
+  // disable save button if not all input fields are filled
+  static getDerivedStateFromProps(props, state) {
+    const values = pick(state, 
+      ['academicLevel', 'choreographers', 'competitionLevel', 'danceEntry',
```
YangStephen: For these inline arrays, let's extract this into another const.

src/components/performances/PerformanceForm.jsx
```diff
@@ -19,20 +20,28 @@ class PerformanceForm extends React.Component {
...
       danceTitle: currentValues.danceTitle || '',
-      disabled: false,
       options: {},
       performers: currentValues.performers || '',
       school: currentValues.school || '',
       size: currentValues.size || 0
     };
   }
 
+  // disable save button if not all input fields are filled
```
YangStephen: NIT: Proper capitalization in comments + a period at the end please!

src/components/performances/PerformanceForm.jsx
```diff
@@ -19,20 +20,28 @@ class PerformanceForm extends React.Component {
...
+  static getDerivedStateFromProps(props, state) {
```
YangStephen: I'll circle back on this in person but I'm not sure if `getDerivedStateFromProps` is the best place to do this. We're actually not deriving the state from the props but instead from the state only.

src/components/performances/PerformanceForm.jsx
meganniu: Reordered with `type` first

src/components/performances/PerformanceForm.jsx
meganniu: Done

src/components/performances/PerformanceForm.jsx
meganniu: Extracted into constant `fields`

src/components/performances/PerformanceForm.jsx
meganniu: Done

src/components/performances/PerformanceForm.jsx
YangStephen: Keeping based on existing precedence, moving this to componentDidUpdate will cause double renders.

PR #52, opened 2019-02-05
Changed filter names

JasonFok:


src/components/interface/filter/FilterMenu.jsx
```diff
@@ -40,10 +40,12 @@ class FilterMenu extends React.Component {
     const { handleFilterChecked } = this.props;
     const choices = this.aggregateChoices(name, options[name]);
     const key = `${name}-${index}`;
+    const result = name.replace( /([A-Z])/g, " $1" );
```
YangStephen: One quick comment, let's rename these variables as something more specific. Like `spacedOutName`, `capitalizedSecondWord`

PR #53, opened 2019-02-06
Placeholder text for date field on event form

meganniu:
Event form now has placeholder text specifying the input format ("DD/MM/YYYY") in the event date field.

### DIFF
--- a/src/components/interface/dialog/DialogInput.jsx
+++ b/src/components/interface/dialog/DialogInput.jsx
@@ -12,6 +12,7 @@ const DialogInput = ({
   label,
   multiline = false,
   onChange,
+  placeholder,
   style,
   type,
   value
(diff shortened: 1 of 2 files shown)


YangStephen:
@meganniu If we're adding in the Calendar dropdown, is this still required? Will we still be using a `DialogInput` or will we be switching to another calendar class that we'll import?

src/components/interface/dialog/DialogInput.jsx
```diff
@@ -57,6 +60,7 @@ DialogInput.defaultProps = {
   multiline: false,
   label: '',
   onChange: () => {},
+  placeholder: '',
```
YangStephen: `DD/MM/YYYY` as a default prop

YangStephen:
Merging for Howard's Demo tomorrow.

PR #54, opened 2019-02-10
Filter award considerations

livictoria:
* Adds ability to filter award considerations
* updates all `new` and `edit` strings into constants
* adds `this.subscibe()` to `componentWillUmount` to handle memory leaks with react state changes

### DIFF
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import isEqual from 'lodash';
+import isEqual from 'lodash/isEqual';
 
 import { withStyles } from '@material-ui/core/styles';
 import DialogActions from '@material-ui/core/DialogActions';
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -1,9 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import isEqual from 'lodash';
+
 import { withStyles } from '@material-ui/core/styles';
 import DialogActions from '@material-ui/core/DialogActions';
 
+import batchUpdateData from '../../firebase/utils/batchUpdateData';
 import updateData from '../../firebase/utils/updateData';
+import { awardConsiderationEnum } from '../../constants';
 
 import DialogInput from '../interface/dialog/DialogInput';
 import Button from '../interface/Button';
@@ -46,18 +50,37 @@ class AdjudicationForm extends React.Component {
     }
   }
 
+  getAwardConsderationEnumAction = (currAwardState, prevAwardState) => {
+    if (!isEqual(currAwardState, prevAwardState)) {
+      return currAwardState ? awardConsiderationEnum.INCREMENT : awardConsiderationEnum.DECREMENT;
+    }
+    return awardConsiderationEnum.NO_CHANGE;
+  }
+
+  // TODO: handle submmission of the form
   handleSubmit = async () => {
-    const { 
-      adjudicationId, 
-      collectionName, 
-      currentValues : {
-        audioURL, 
-        judgeName, 
-        notes
-      } 
+    const {
+      adjudicationId,
+      collectionName,
+      currentValues: {
+        audioURL,
+        choreoAward: prevChoreoAward,
+        judgeName,
+        notes,
+        specialAward: prevSpecialAward
+      }
     } = this.props;
-    const { artisticMark, technicalMark } = this.state;
+    const {
+      artisticMark,
+      choreoAward,
+      specialAward,
+      technicalMark
+    } = this.state;
     const cumulativeMark = (parseInt(artisticMark, 10) + parseInt(technicalMark, 10)) / 2;
+
+    const choreoAwardAction = this.getAwardConsderationEnumAction(choreoAward, prevChoreoAward);
+    const specialAwardAction = this.getAwardConsderationEnumAction(specialAward, prevSpecialAward);
+
     const data = {
       audioURL,
       cumulativeMark,
@@ -65,11 +88,23 @@ class AdjudicationForm extends React.Component {
       notes,
       ...this.state
     };
-    await updateData(
-      collectionName,
-      adjudicationId,
-      data
-    );
+
+    if (choreoAwardAction !== awardConsiderationEnum.NO_CHANGE
+      || specialAwardAction !== awardConsiderationEnum.NO_CHANGE) {
+      await batchUpdateData(
+        collectionName,
+        adjudicationId,
+        data,
+        choreoAwardAction,
+        specialAwardAction
+      );
+    } else {
+      await updateData(
+        collectionName,
+        adjudicationId,
+        data,
+      );
+    }
     this.handleModalClose();
   }
 
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -83,12 +83,15 @@ class AdjudicationForm extends React.Component {
     const specialAwardAction = this.getAwardConsderationEnumAction(specialAward, prevSpecialAward);
 
     const data = {
-      audioURL,
       cumulativeMark,
-      judgeName,
+      judgeName: judgeName || '',
       notes,
       ...this.state
     };
+
+    if (audioURL) {
+      Object.assign(data, audioURL);
+    }
 
     const { NO_CHANGE } = awardConsiderationEnum;
 
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -51,10 +51,11 @@ class AdjudicationForm extends React.Component {
   }
 
   getAwardConsderationEnumAction = (currAwardState, prevAwardState) => {
+    const { DECREMENT, INCREMENT, NO_CHANGE } = awardConsiderationEnum;
     if (!isEqual(currAwardState, prevAwardState)) {
-      return currAwardState ? awardConsiderationEnum.INCREMENT : awardConsiderationEnum.DECREMENT;
+      return currAwardState ? INCREMENT : DECREMENT;
     }
-    return awardConsiderationEnum.NO_CHANGE;
+    return NO_CHANGE;
   }
 
   // TODO: handle submmission of the form
@@ -89,8 +90,10 @@ class AdjudicationForm extends React.Component {
       ...this.state
     };
 
-    if (choreoAwardAction !== awardConsiderationEnum.NO_CHANGE
-      || specialAwardAction !== awardConsiderationEnum.NO_CHANGE) {
+    const { NO_CHANGE } = awardConsiderationEnum;
+
+    if (choreoAwardAction !== NO_CHANGE
+      || specialAwardAction !== NO_CHANGE) {
       await batchUpdateData(
         collectionName,
         adjudicationId,
--- a/src/components/events/EventDialog.jsx
+++ b/src/components/events/EventDialog.jsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
 import CalendarTodayIcon from '@material-ui/icons/CalendarToday';
 
+import { dialogType } from '../../constants';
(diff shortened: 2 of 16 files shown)


src/components/events/EventDialog.jsx
```diff
@@ -31,13 +32,13 @@ class EventDialog extends React.Component {
   render() {
     const { currentValues, eventId, formType } = this.props;
     const { open } = this.state;
-    const dialogTitle = formType === 'edit' ? 'Edit' : 'New';
+    const dialogTitle = formType === dialogType.EDIT ? 'Edit' : 'New';
```
YangStephen: Can we destructure these references of dialogType?

src/components/interface/editing/AlertDialog.jsx
```diff
@@ -19,7 +19,7 @@ const AlertDialog = ({
   const handleClose = () => onClose();
 
   const handleDelete = async () => {
-    await updateSettings(category, optionName, constants.DELETE_DATA);
+    await updateSettings(category, optionName, dataAction.DELETE_DATA);
```
YangStephen: destructure here please :)

src/components/performances/PerformancesSection.jsx
```diff
@@ -70,10 +73,24 @@ class PerformancesSection extends React.Component {
     // Filter function that accepts a performance object and returns true
     // if this performance passes the filter (i.e should be included)
     const filterFunction = (performance) => {
-      const danceTitle = performance.danceTitle.toLowerCase();
+      const danceTitle = performance.danceTitle ? performance.danceTitle.toLowerCase() : '';
       // Iterates through each performance metadata/key (i.e. academicLevel) and then
       // iterates through each string in the array (i.e. secondary, primary)
-      const isFilterSuccess = keys.every(key => filtersObj[key].includes(performance[key]));
+      const isFilterSuccess = keys.every((key) => {
+        // awardConiderations need to be checked differently as it's a Number rather
```
YangStephen: `awardConsiderations` **

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,57 @@
+import db from '../firebase';
+import { awardConsiderationEnum } from '../../constants';
+
+/*
+* This batchUpdateData method is written explicitly for handling
```
YangStephen: Let's space this such that all the asterisks are in line with the asterisk on line 4.

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,57 @@
...
+const batchUpdateData = async (
+  collectionName,
+  docName,
+  data,
+  choreoAwardAction,
+  specialAwardAction,
+) => {
+  const performanceCollection = collectionName.match(/events\/\w{20}\/performances/)[0];
+  const { length } = performanceCollection;
+  const performanceDocName = collectionName.substring(length + 1, length + 21);
```
YangStephen: What is encapsulated by the substring at length + 1 and length + 21. Is this a performanceDocName or performanceDocID then?

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const ajudicationRef = db.collection(collectionName).doc(docName);
```
YangStephen: `adjudication`**

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,57 @@
...
+  const performanceRef = db.collection(performanceCollection).doc(performanceDocName);
+
+  db.runTransaction(transaction => transaction.get(performanceRef).then((performanceDoc) => {
```
YangStephen: This potentially breaks the 1 line char limit, let's space this out into multiple lines if possible

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,57 @@
...
+    const awardData = {};
+
+    if (choreoAwardAction !== awardConsiderationEnum.NO_CHANGE) {
+      let idx;
```
YangStephen: could lines 27-32, and lines 38 - 43 be consolidated into a helper function? One option is to create the helper function within this function so it has access to performanceDoc or to create it outside and pass in performanceDoc.

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,57 @@
...
+  const { length } = performanceCollection;
+  const performanceDocName = collectionName.substring(length + 1, length + 21);
+
+  const ajudicationRef = db.collection(collectionName).doc(docName);
+  const performanceRef = db.collection(performanceCollection).doc(performanceDocName);
+
+  db.runTransaction(transaction => transaction.get(performanceRef).then((performanceDoc) => {
+    const awardData = {};
+
+    if (choreoAwardAction !== awardConsiderationEnum.NO_CHANGE) {
```
YangStephen: For the uses of the awardConsiderationEnum, could we destructure NO_CHANGE and DECREMENT?

src/components/adjudications/AdjudicationForm.jsx
```diff
@@ -46,20 +50,55 @@ class AdjudicationForm extends React.Component {
     }
   }
 
+  getAwardConsderationEnumAction = (currAwardState, prevAwardState) => {
+    if (!isEqual(currAwardState, prevAwardState)) {
+      return currAwardState ? awardConsiderationEnum.INCREMENT : awardConsiderationEnum.DECREMENT;
```
YangStephen: Do you think we should destructure all these references of awardConsiderationEnum? @livictoria

src/components/performances/PerformanceDialog.jsx
```diff
@@ -31,9 +32,9 @@ class PerformanceDialog extends React.Component {
     } = this.props;
     const { open } = this.state;
     const collectionName = `events/${eventId}/performances`;
-    const shouldShowDropdown = formType === 'edit';
```
YangStephen: I'd really prefer we destructure EDIT here as well since it's used 3 times.

src/components/performances/PerformanceForm.jsx
```diff
@@ -76,7 +77,7 @@ class PerformanceForm extends React.Component {
   handleSubmit = async () => {
     const { collectionName, formType, performanceId } = this.props;
     const data = omit(this.state, ['disabledSave', 'options']);
-    if (formType === 'new') {
+    if (formType === dialogType.NEW) {
```
YangStephen: and destructuring new / edit in this file

src/components/performances/PerformancesSection.jsx
```diff
@@ -71,10 +74,24 @@ class PerformancesSection extends React.Component {
...
       // iterates through each string in the array (i.e. secondary, primary)
-      const isFilterSuccess = keys.every(key => filtersObj[key].includes(performance[key]));
+      const isFilterSuccess = keys.every((key) => {
+        // awardConsiderations need to be checked differently as it's a Number rather
+        // than a String
+        if (key === 'awardConsideration') {
+          if (filtersObj[key].includes('choreoAward')) {
+            return performance.choreoAwardEnum > 0;
+          }
+          if (filtersObj[key].includes('specialAward')) {
```
YangStephen: else if here.

src/components/performances/PerformancesSection.jsx
```diff
@@ -71,10 +74,24 @@ class PerformancesSection extends React.Component {
...
+            return performance.specialAwardEnum > 0;
+          }
+
+          return false;
```
YangStephen: I'm curious if we should be returning false here. This is essentially a case where somehow we have a filter key on awardConsideration but none of the awards are checked. To me, I almost feel we should ignore this awardConsideration if choreoAward and specialAward are both missing.

src/firebase/utils/batchUpdateData.js
```diff
@@ -0,0 +1,68 @@
+import db from '../firebase';
+import { awardConsiderationEnum } from '../../constants';
+
+const getIndex = (action, performanceDocDataEnum) => {
+  const { DECREMENT } = awardConsiderationEnum;
```
YangStephen: I'm curious if we could also get some documentation on this. I'm confused how this IDX logic below works in calculating an index.

src/components/performances/PerformancesSection.jsx
livictoria: @YangStephen  adjudications can have both choreAward and specialAward considerations

src/firebase/utils/batchUpdateData.js
livictoria: @YangStephen  can do this in a separate PR this weekend

src/firebase/utils/batchUpdateData.js
livictoria: adding documentation above L4-35

src/components/performances/PerformancesSection.jsx
livictoria: actually, i guess we're iteration through each key, then yes it should be else if

YangStephen: Yeah, please do verify the my logic is correct. I'd assume if the filterObj contains choreoAward, then choreoAward must be ticked. So it makes sense to forcefully return out of this statement since the failure of choreoAward to match here means that the current performance should not be filtered on.

Else if was just to change the if-return, if-return pattern.

src/components/performances/PerformancesSection.jsx
YangStephen: Just to clarify the above, lets say my filter Object looks like this:

```
{
    awardConsideration : {}
}
```

I'd argue that this award consideration is not a valid award consideration. This would actually cause *all* entries to get kicked out since no filter function on a performance will return true in this case, it would return false.

livictoria: when you mean:

> none of the awards are checked

do you mean that none of the awards have an award consideration?

livictoria: Actually nvm, I understand what you mean. I think this is actually pretty confusing so i'll explain it thoroughly. To clarify (for future reference), we have a `filterObj` that looks like this: 
```
{
  academicLevel: [],
  awardConsideration: [],
  competitionLevel: [],
  danceSize: [],
  danceStyle: [],
  school: [],
}
```

The method iterates through each key in the filterObj, when we get to `awardConsideration`, we check if the array includes `specialAward` and `choreoAward`. In the case where we have an empty array for the value of `awardsConsideration`, if there is no `choreoAward` or `specialAward` checked off to be filter (a.k.a `awardsConsideration: [] `), should we return true or false?

So yea you're right, If there are no filter values of `choreoAward` and `specialAward`, we should be returning `true`, b/c we've never set any filters on awards consideration. 

I tested it out, and technically, both work b/c:

 > The every() method tests whether all elements in the array pass the test implemented by the provided function

and this entire filter logic only gets called if whenever a filter is checked. 
 
ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

src/components/performances/PerformancesSection.jsx
livictoria: okay, I was actually confused above or disregard what I said before. After thinking and testing, both logics work (and I can explain that another time). However, I am leaning towards having it the way it is: 

```
if (filtersObj[key].includes('choreoAward')) {
    return performance.choreoAwardEnum > 0;
}
if (filtersObj[key].includes('specialAward')) {
    return performance.specialAwardEnum > 0;
}

return true;
```

why? Because the result is not conditional, this is an example of what `filterObj` looks like with values in `awardConsideration`:

``
{
   awardConsideration : ["choreoAward", "specialAward"]
}
```

so when `key === "awardConsideration`, we could have both, so technically, 
if (.... ) else if (...) else (...) doesn't really match the logic we're trying to portray. B/c you can have a `filterObjc[key]` that has both. Hope this helps! :)

YangStephen:
I've tested this and it works really well! Thanks @livictoria :)

src/components/performances/PerformancesSection.jsx
YangStephen: This will be fine :)

src/components/performances/PerformancesSection.jsx
YangStephen: Got it, I won't debate too much on this topic but my main reason is that *from a code perspective* the second if cannot execute without the first.

PR #55, opened 2019-02-12
Made settings category dropdown human readable

JasonFok:


YangStephen:
Let's hold off on this PR. I'd like to thoroughly test this with you in person as we have shown that this is potentially a breaking change.

src/components/settings/SettingsSection.jsx
```diff
@@ -40,7 +42,8 @@ class SettingsSection extends React.Component {
 
   handleSubmit = async () => {
     const { selectedCategory, value } = this.state;
-    await updateSettings(selectedCategory, value, constants.ADD_DATA);
+    const selectedCategoryCamelCase = selectedCategory.charAt(0).toLowerCase() + selectedCategory.replace(/ /g,'').slice(1);
```
YangStephen: This code is used twice, can we get a helper function to handle this instead?

PR #60, opened 2019-03-02
Fix search

wrjren:
Search checks for the existence of the keyword in the dance title, performers, and choreographers.


meganniu:
@YangStephen For sure :)

PR #63, opened 2019-03-08
Dropdown calendar for Event Form

meganniu:
Dropdown calendar implemented for date field of Event Form. If a date exists, the calendar auto-focuses on the date upon opening. If no date exists, default date is the current date.

### DIFF
--- a/src/components/events/EventForm.jsx
+++ b/src/components/events/EventForm.jsx
@@ -35,7 +35,8 @@ class EventForm extends React.Component {
 
   handleChange = (e) => {
     const { name, value } = e.target;
-    this.setState({ [name]: name === 'eventDate' ? (value.search(/-/) === -1 ? moment(value, 'MM/DD/YYYY').format('DD/MM/YYYY') : moment(value, 'YYYY-MM-DD').format('DD/MM/YYYY')) : value });
+    console.log(value);
+    this.setState({ [name]: name === 'eventDate' && value ? (value.search(/-/) === -1 ? moment(value, 'MM/DD/YYYY').format('DD/MM/YYYY') : moment(value, 'YYYY-MM-DD').format('DD/MM/YYYY')) : value });
   }
 
   handleCancel = () => {
--- a/src/components/events/EventForm.jsx
+++ b/src/components/events/EventForm.jsx
@@ -23,7 +23,6 @@ class EventForm extends React.Component {
       eventDate: currentValues.eventDate || moment().format('DD/MM/YYYY'),
       eventTitle: currentValues.eventTitle || '',
       numJudges: currentValues.numJudges || 0,
-      focused: false
     };
   }
 
--- a/src/components/events/EventForm.jsx
+++ b/src/components/events/EventForm.jsx
@@ -79,7 +79,7 @@ class EventForm extends React.Component {
         <div style={{ margin: '25px' }}>
           <DialogInput fullWidth name="eventTitle" label="Event Title" onChange={this.handleChange} value={eventTitle} />
           <div style={{ display: 'flex' }}>
-            <DialogInput fullWidth name='eventDate' label='Event Date' onChange={this.handleChange} style={{ marginRight: '5px' }} type="date" value={eventDate ? moment(eventDate, 'DD/MM/YYYY').format('YYYY-MM-DD') : ''} variant="filled" />
+            <DialogInput fullWidth InputLabelProps={{ shrink: true }} name='eventDate' label='Event Date' onChange={this.handleChange} style={{ marginRight: '5px' }} type="date" value={eventDate ? moment(eventDate, 'DD/MM/YYYY').format('YYYY-MM-DD') : ''} variant="filled" />
             <DialogInput fullWidth name="numJudges" label="No. Judges" onChange={this.handleChange} type="number" value={numJudges} />
           </div>
         </div>
--- a/src/components/events/EventForm.jsx
+++ b/src/components/events/EventForm.jsx
@@ -35,7 +35,7 @@ class EventForm extends React.Component {
 
   handleChange = (e) => {
     const { name, value } = e.target;
-    // If name is 'eventDate', then the format of value is 'DD/MM/YYYY'.
+    // If name is 'eventDate', then the format of value is 'YYYY-MM-DD'.
     this.setState({ [name]: name === 'eventDate' ? value.replace(/(\d{4})\-(\d{2})\-(\d{2})/, '$3/$2/$1') : value });
   }
 
--- a/src/components/events/EventForm.jsx
+++ b/src/components/events/EventForm.jsx
@@ -36,7 +36,12 @@ class EventForm extends React.Component {
   handleChange = (e) => {
     const { name, value } = e.target;
     console.log(value);
-    this.setState({ [name]: name === 'eventDate' && value ? (value.search(/-/) === -1 ? moment(value, 'MM/DD/YYYY').format('DD/MM/YYYY') : moment(value, 'YYYY-MM-DD').format('DD/MM/YYYY')) : value });
+    if(name === 'eventDate' && value){
+      this.setState({ [name]: value.search(/-/) === -1 ? moment(value, 'MM/DD/YYYY').format('DD/MM/YYYY') : moment(value, 'YYYY-MM-DD').format('DD/MM/YYYY') });
+    }
+    else{
+      this.setState({ [name]: value });
+    }
   }
 
   handleCancel = () => {
@@ -79,7 +84,17 @@ class EventForm extends React.Component {
         <div style={{ margin: '25px' }}>
           <DialogInput fullWidth name="eventTitle" label="Event Title" onChange={this.handleChange} value={eventTitle} />
           <div style={{ display: 'flex' }}>
-            <DialogInput fullWidth InputLabelProps={{ shrink: true }} name='eventDate' label='Event Date' onChange={this.handleChange} style={{ marginRight: '5px' }} type="date" value={eventDate ? moment(eventDate, 'DD/MM/YYYY').format('YYYY-MM-DD') : ''} variant="filled" />
+            <DialogInput 
+              fullWidth 
+              InputLabelProps={{ shrink: true }} 
+              name='eventDate' 
+              label='Event Date' 
+              onChange={this.handleChange} 
+              style={{ marginRight: '5px' }} 
+              type="date" 
+              value={eventDate ? moment(eventDate, 'DD/MM/YYYY').format('YYYY-MM-DD') : ''} 
+              variant="filled"
+            />
             <DialogInput fullWidth name="numJudges" label="No. Judges" onChange={this.handleChange} type="number" value={numJudges} />
           </div>
         </div>
--- a/src/components/events/EventForm.jsx
+++ b/src/components/events/EventForm.jsx
@@ -10,7 +10,8 @@ import updateData from '../../firebase/utils/updateData';
 
 import DialogInput from '../interface/dialog/DialogInput';
 import Button from '../interface/Button';
-import styles from '../styles';
+import styles from '../styles'; 
+import TextField from '@material-ui/core/TextField';
 
 class EventForm extends React.Component {
   constructor(props) {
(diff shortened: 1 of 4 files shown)


meganniu:
screenshot from 2019-03-07 22-45-52

src/components/events/EventForm.jsx
```diff
@@ -10,7 +10,8 @@ import updateData from '../../firebase/utils/updateData';
 
 import DialogInput from '../interface/dialog/DialogInput';
 import Button from '../interface/Button';
-import styles from '../styles';
+import styles from '../styles';
```
YangStephen: This line doesn't look like it's changed, what's the diff from here?

src/components/events/EventForm.jsx
```diff
@@ -10,7 +10,8 @@ import updateData from '../../firebase/utils/updateData';
 
 import DialogInput from '../interface/dialog/DialogInput';
 import Button from '../interface/Button';
-import styles from '../styles';
+import styles from '../styles'; 
+import TextField from '@material-ui/core/TextField';
```
YangStephen: This Material UI should be imported with the group on lines 5/6

src/components/events/EventForm.jsx
```diff
@@ -33,7 +35,12 @@ class EventForm extends React.Component {
 
   handleChange = (e) => {
     const { name, value } = e.target;
-    this.setState({ [name]: value });
+    if (name === 'eventDate') {
+      this.setState({ eventDate: value.substring(8,10) + '/' + value.substring(5,7) + '/' + value.substring(0,4)});
```
YangStephen: What's the format that these values are coming in as? This Substring function seems to be assuming formatting of the input.

src/components/events/EventForm.jsx
```diff
@@ -33,7 +35,12 @@ class EventForm extends React.Component {
...
+    }
+    else {
+      this.setState({ [name]: value });
```
YangStephen: There's a fun way we could get lines 38 - 43 condensed into one line, theoretically.

If i run this `this.setState({ [name]: value })` when name is eventDate, the key of the object will still be valid. So we could actually condense both the setState on lines 39 and 42 into one setState function.

```
this.setState({ [name]: name === 'eventDate' ? value...... : value });
```

src/components/events/EventForm.jsx
```diff
@@ -46,7 +53,7 @@ class EventForm extends React.Component {
     const { eventDate, eventTitle, numJudges } = this.state;
     const collectionName = 'events';
     const data = {
-      eventDate: new Date(eventDate),
+      eventDate: new Date(parseInt(eventDate.substring(6, 10)), parseInt(eventDate.substring(3, 5)) - 1, parseInt(eventDate.substring(0, 2)), 4, 0, 0),
```
YangStephen: can the constructor for date accept a locale string?

src/components/events/EventForm.jsx
```diff
@@ -63,6 +70,12 @@ class EventForm extends React.Component {
     onModalClose();
   }
 
+  //Format of date must be DD/MM/YYYY
+  //Transforms DD/MM/YYYY to YYYY-MM-DD
+  convertDateToISOFormat = (date) => {
+    return date.substring(6, 10) + "-" + date.substring(3, 5) + "-" + date.substring(0, 2);
```
YangStephen: I'd prefer we don't use the substring formatting approach if possible. Typically, date objects have a way where you can call `toString` or a similar function and pass in a string format like: `MM/DD/YYYY` which automatically parses to `01/01/1993`. Could we explore if that is possible?

src/components/events/EventForm.jsx
```diff
@@ -63,6 +70,12 @@ class EventForm extends React.Component {
     onModalClose();
   }
 
+  //Format of date must be DD/MM/YYYY
```
YangStephen: When we do these function comments, could we phrase all statements in full sentences?

src/components/events/EventForm.jsx
```diff
@@ -76,7 +89,7 @@ class EventForm extends React.Component {
         <div style={{ margin: '25px' }}>
           <DialogInput fullWidth name="eventTitle" label="Event Title" onChange={this.handleChange} value={eventTitle} />
           <div style={{ display: 'flex' }}>
-            <DialogInput style={{ marginRight: '5px' }} fullWidth name="eventDate" label="Event Date" onChange={this.handleChange} placeholder="DD/MM/YYYY" value={eventDate} />
+            <TextField  fullWidth name='eventDate' label='Event Date' onChange={this.handleChange} style={{ marginRight: '5px' }} type="date" value={eventDate ? this.convertDateToISOFormat(eventDate) : ''} variant="filled" />
```
YangStephen: Why do we need to convert this from a DialogInput to a TextField? Is a DialogInput not our implementation of the MaterialUI Textfield?

src/components/events/EventForm.jsx
meganniu: That's really weird, they're exactly identical.

src/components/events/EventForm.jsx
meganniu: The formatting in the textfield ensures that the format of value is DD/MM/YYYY. I have added a comment above line 39 to inform future developers of the format of value.

src/components/events/EventForm.jsx
meganniu: Once again, you're a genius.

src/components/events/EventForm.jsx
meganniu: Yes, the date constructor can accept a string. There was a problem where the date was assumed to be EDT when passed in and when the date object was translated into a string, it would produce UTC which is four hours behind. So if I passed in 20/02/2000, then the date objects would actually be 4 hours behind of 20/02/2000 00:00 EDT and the date represented would be 19/20/2000 17:00 UTC. I fixed this by adding 4 hours to 00:00 EDT so that when the date is converted to UTC, the date represented would still be the same intended date. 

But yes, the solution is very inelegant. I think I found a better solution for this and hopefully it works.

src/components/events/EventForm.jsx
YangStephen: Preferably we leave lines we don't touch alone. If we need Git Blame later on, it's hard to see who the original author / modifier of this line was since the latest git change would have your signature on it.

YangStephen:
Nice job Meg! i have a few things i'll talk to you about in person, one quick note however.

There's a commit that you've merged in from fixSearch that's showing us modifying Victoria's search function. Let's see if we can revert this commit. I thought that was already merged into master but let's keep that separate from this.

The commit ID is: d124a3f2822721822cde4fe3d6b5b361edef7326

src/components/events/EventForm.jsx
YangStephen: It looks like theres a space on the diff here?

src/components/events/EventForm.jsx
```diff
@@ -19,9 +20,11 @@ class EventForm extends React.Component {
 
     this.state = {
       disabledSave: true,
-      eventDate: currentValues.eventDate || '',
+      //eventDate: currentValues.eventDate || (new Date()).toLocaleDateString('en-US'),
```
YangStephen: why is this commented out?

src/components/events/EventsSection.jsx
```diff
@@ -23,8 +23,9 @@ class EventsSection extends React.Component {
       querySnapshot.forEach((doc) => {
         const { eventTitle, numJudges, eventDate } = doc.data();
         const date = isObject(eventDate)
-          ? new Date(eventDate.seconds * 1000).toLocaleDateString()
+          ? new Date(eventDate.seconds * 1000).toLocaleDateString('en-GB')
           : eventDate;
+          console.log(date);
```
YangStephen: extra cosnole log.

src/components/events/EventsSection.jsx
```diff
@@ -23,8 +23,9 @@ class EventsSection extends React.Component {
       querySnapshot.forEach((doc) => {
         const { eventTitle, numJudges, eventDate } = doc.data();
         const date = isObject(eventDate)
-          ? new Date(eventDate.seconds * 1000).toLocaleDateString()
+          ? new Date(eventDate.seconds * 1000).toLocaleDateString('en-GB')
```
YangStephen: what is this en-GB?

src/components/events/EventForm.jsx
```diff
@@ -76,7 +79,7 @@ class EventForm extends React.Component {
         <div style={{ margin: '25px' }}>
           <DialogInput fullWidth name="eventTitle" label="Event Title" onChange={this.handleChange} value={eventTitle} />
           <div style={{ display: 'flex' }}>
-            <DialogInput style={{ marginRight: '5px' }} fullWidth name="eventDate" label="Event Date" onChange={this.handleChange} placeholder="DD/MM/YYYY" value={eventDate} />
+          <DialogInput fullWidth name='eventDate' label='Event Date' onChange={this.handleChange} style={{ marginRight: '5px' }} type="date" value={eventDate ? moment(eventDate, 'DD/MM/YYYY').format('YYYY-MM-DD') : ''} variant="filled" />
```
YangStephen: Why did we change the indentation of this line? And is it ok to make this value an empty string when eventDate is non-truthy?

src/components/events/EventForm.jsx
meganniu: Sorry, forgot to delete the old line of code

src/components/events/EventsSection.jsx
meganniu: Formats date object as DD/MM/YYYY

src/components/events/EventForm.jsx
```diff
@@ -33,7 +35,13 @@ class EventForm extends React.Component {
 
   handleChange = (e) => {
     const { name, value } = e.target;
-    this.setState({ [name]: value });
+    console.log(value);
+    if(name === 'eventDate' && value){
+      this.setState({ [name]: value.search(/-/) === -1 ? moment(value, 'MM/DD/YYYY').format('DD/MM/YYYY') : moment(value, 'YYYY-MM-DD').format('DD/MM/YYYY') });
```
YangStephen: wouldn't includes make more sense than search here? https://www.w3schools.com/jsref/jsref_includes.asp

YangStephen: You could segment thiis into multiple lines for readability

```
this.setState({
   [name] .....
});
```

Typically we aim for ~100 characters a line as a max line length. Anything past that and you should start playing around with how the code is formatted.

src/components/events/EventForm.jsx
```diff
@@ -76,7 +84,17 @@ class EventForm extends React.Component {
         <div style={{ margin: '25px' }}>
           <DialogInput fullWidth name="eventTitle" label="Event Title" onChange={this.handleChange} value={eventTitle} />
           <div style={{ display: 'flex' }}>
-            <DialogInput style={{ marginRight: '5px' }} fullWidth name="eventDate" label="Event Date" onChange={this.handleChange} placeholder="DD/MM/YYYY" value={eventDate} />
+            <DialogInput 
+              fullWidth 
+              InputLabelProps={{ shrink: true }}
```
YangStephen: what is shrink?

src/components/interface/dialog/DialogInput.jsx
```diff
@@ -25,6 +26,7 @@ const DialogInput = ({
     disabled={disabled}
     style={{ margin: '0 5px 25px 5px', ...style }}
     helperText={helperText}
+    InputLabelProps = {InputLabelProps}
```
YangStephen: Why do we need these InputLabelProps?

Also remember that we don't need to follow the same syntax, we can convert ours to follow camelCase style.

This function could take in `inputLabelProps` and say `InputLabelProps={inputLabelProps}`

src/components/events/EventForm.jsx
meganniu: Yup!

src/components/events/EventForm.jsx
meganniu: `shrink` is being used to fix the problem where the label for date text field would overlap with the placeholder when the field was not focused.

Some more info: https://material-ui.com/demos/text-fields/#limitations

src/components/interface/dialog/DialogInput.jsx
meganniu: Changed to camel case.

I added `InputLabelProps` to solve the overlap between the text field label and the value in the text field when the text field is not focused.

PR #64, opened 2019-03-12
Fix for Adjudication Form sumbit bug

meganniu:
Updating Firebase from adjudication form no longer attempts to create new fields with undefined values.

### DIFF
--- a/src/components/adjudications/AdjudicationForm.jsx
+++ b/src/components/adjudications/AdjudicationForm.jsx
@@ -50,19 +50,14 @@ class AdjudicationForm extends React.Component {
     const { 
       adjudicationId, 
       collectionName, 
-      currentValues : {
-        audioURL, 
-        judgeName, 
-        notes
-      } 
+      currentValues
     } = this.props;
     const { artisticMark, technicalMark } = this.state;
     const cumulativeMark = (parseInt(artisticMark, 10) + parseInt(technicalMark, 10)) / 2;
+
     const data = {
-      audioURL,
       cumulativeMark,
-      judgeName,
-      notes,
+      ...currentValues,
       ...this.state
     };
     await updateData(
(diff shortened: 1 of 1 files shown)


src/components/adjudications/AdjudicationForm.jsx
```diff
@@ -50,19 +50,14 @@ class AdjudicationForm extends React.Component {
...
-      currentValues : {
-        audioURL, 
-        judgeName, 
-        notes
-      } 
+      currentValues
     } = this.props;
     const { artisticMark, technicalMark } = this.state;
     const cumulativeMark = (parseInt(artisticMark, 10) + parseInt(technicalMark, 10)) / 2;
+
```
YangStephen: Let's remove this extra line

src/components/adjudications/AdjudicationForm.jsx
```diff
@@ -50,19 +50,14 @@ class AdjudicationForm extends React.Component {
...
     const data = {
-      audioURL,
       cumulativeMark,
-      judgeName,
-      notes,
+      ...currentValues,
```
YangStephen: So the PR comment says this will restrict undefined values from being sent to firebase. Is it possible that currentValues will have something as `undefined`?

meganniu: I don't believe that `currentValues` will hold any undefined values.

`currentValues` is created in AdjudicationsSection.jsx and passed down through several children until it reaches AdjudicationForm.jsx. I traced the contents of currentValues as it is being passed and it seems that if a field is not defined in Firebase, it will not appear in `currentValues`.

src/components/adjudications/AdjudicationForm.jsx
```diff
@@ -50,19 +50,14 @@ class AdjudicationForm extends React.Component {
...
-        notes
-      } 
+      currentValues
     } = this.props;
     const { artisticMark, technicalMark } = this.state;
     const cumulativeMark = (parseInt(artisticMark, 10) + parseInt(technicalMark, 10)) / 2;
+
     const data = {
-      audioURL,
       cumulativeMark,
```
YangStephen: If i recall correctly, I had a bug running taking this spread object. 

The value of cumulativeMark in this object comes from the value of this.state, then currentValues, then this cumulativeMark value. The way the spread operator works is you're defining an object, saying add and overwrite everything so far with currentValues, then add/overwrite with this.state. 

We need to move cumulativeMark to the bottom of this object declaration.

PR #65, opened 2019-03-12
Display performance fields in Adjudication Dialog 

meganniu:
Adjudication dialog now displays details of performance associated with the adjudication (dance title, performers, choreographers, school etc.)

### DIFF
--- a/src/components/adjudications/AdjudicationDialog.jsx
+++ b/src/components/adjudications/AdjudicationDialog.jsx
@@ -31,7 +31,12 @@ class AdjudicationDialog extends React.Component {
   }
 
   render() {
-    const { adjudicationId, collectionName, currentValues, performanceValues } = this.props;
+    const {
+      adjudicationId,
+      collectionName,
+      currentValues,
+      performanceValues
+    } = this.props;
     const { open, view } = this.state;
     const {
       artisticMark,
@@ -43,10 +48,10 @@ class AdjudicationDialog extends React.Component {
       specialAward,
       technicalMark
     } = currentValues;
-
     const {
       academicLevel,
       choreographers,
+      competitionLevel,
       danceEntry,
       danceStyle,
       danceTitle,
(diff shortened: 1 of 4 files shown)


src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -44,25 +44,38 @@ class AdjudicationDialog extends React.Component {
...
+      choreographers,
+      danceEntry,
+      danceStyle,
+      danceTitle,
+      performers,
+      school,
+      size
+    } = performanceValues;
+
+    console.log(performanceValues);
```
YangStephen: stray console log

src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -31,7 +31,7 @@ class AdjudicationDialog extends React.Component {
   }
 
   render() {
-    const { adjudicationId, collectionName, currentValues } = this.props;
+    const { adjudicationId, collectionName, currentValues, performanceValues } = this.props;
```
YangStephen: let's put each of these parameters on a new line.

src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -44,25 +44,38 @@ class AdjudicationDialog extends React.Component {
...
+              <DialogReadOnly defaultValue={danceTitle ? danceTitle : "N/A"} label="Dance Title" />
+              <DialogReadOnly defaultValue={performers ? performers : "N/A"} label="Performers" />
+              <DialogReadOnly defaultValue={choreographers ? choreographers : "N/A"} label="Choreographer" />
             </div>
 
             <div style={{ flex: '1 0 0' }}>
               <div style={{ display: 'flex' }}>
-                <DialogReadOnly label="School" />
-                <DialogReadOnly label="Level" />
+                <DialogReadOnly defaultValue={school ? school : "N/A"} label="School" />
```
YangStephen: do we want this to be "defaultValue" or  is there a value parameter?

The use of school ? school : "N/A" means if the first exists, let's take that value, else "N/A". This is logically equivalent to the Elvis operator from Java `A ?: B` which takes A if it is truthy, else B. The javascript equivalent is `A || B`. It will take A if it exists, else it will return B.

Ref: https://stackoverflow.com/questions/6613952/is-there-a-null-coalescing-elvis-operator-or-safe-navigation-operator-in-javas

Also, All of these are the exact same check, maybe we should add an inline function to handle this?

```
const handleInvalidData = (str) => str || "N/A";
...
defaultValue ={handleInvalidData(school / performers / etc.)}
```

src/components/adjudications/AdjudicationDialog.jsx
```diff
@@ -131,7 +144,8 @@ AdjudicationDialog.propTypes = {
     judgeName: PropTypes.string,
     specialAward: PropTypes.bool,
     technicalMark: PropTypes.number
-  }).isRequired
+  }).isRequired,
+  performanceValues: PropTypes.shape().isRequired
```
YangStephen: Typically when we add in a `PropTypes.shape()`, we also pass in a structure to assure that our shape is valid.

Ex: https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes Look at optionalObjectWithShape

src/components/adjudications/AdjudicationTableRow.jsx
```diff
@@ -58,7 +60,8 @@ AdjudicationTableRow.propTypes = {
     technicalMark: PropTypes.number
   }).isRequired,
   collectionName: PropTypes.string.isRequired,
-  id: PropTypes.string.isRequired
+  id: PropTypes.string.isRequired,
+  performanceValues: PropTypes.shape().isRequired
```
YangStephen: Same here

src/components/adjudications/AdjudicationsSection.jsx
```diff
@@ -12,12 +12,19 @@ class AdjudicationsSection extends React.Component {
...
-      loading: true
+      loading: true,
+      performanceValues: {}
     };
   }
 
   componentDidMount() {
     const { match: { params: { eventId, performanceId }}} = this.props;
+
+    const docRef = db.collection('events').doc(eventId).collection('performances').doc(performanceId);
```
YangStephen: Is it possible for us to use onSnapshot rather than get() here? Get is a one time request, onSnapshot registers a callback to continually update state if the performance changes. 

Also: The collection.doc.collection.doc... call can be summarized. Take a look at the collectionName and the db call below. It summarizes all the collections and docs into one path and then expands on that.

PR #66, opened 2019-03-15
Setup Firebase Deploy, Fix filtering on size

YangStephen:
Filtering on size doesn't work because there's a discrepancy between "size" in our data model and "danceSize". So this converts everything to be under "danceSize"

Additionally: I've included a `firebase.json` which sets up `firebase deploy` from the command line allowing us to deploy an instance to firebase hosting.

### DIFF
--- a/src/components/adjudications/AdjudicationsSection.jsx
+++ b/src/components/adjudications/AdjudicationsSection.jsx
@@ -38,9 +38,8 @@ class AdjudicationsSection extends React.Component {
     const { adjudications, loading } = this.state;
     const collectionName = `events/${eventId}/performances/${performanceId}/adjudications`;
     const headings = ['Judge', 'Audio', 'Cumulative Score', 'Awards'];
-    const keys = ['artisticMark', 'audioURL', 'choreoAward', 'cumulativeMark', 'judgeName', 'notes', 'specialAward', 'technicalMark'];
+    const keys = ['artisticMark', 'audioURL', 'choreoAward', 'cumulativeMark', 'notes', 'specialAward', 'technicalMark', 'tabletID'];
     const showAdjudications = Array.isArray(adjudications) && adjudications.length > 0;
-
     return (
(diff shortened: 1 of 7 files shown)


src/components/adjudications/AdjudicationsSection.jsx
```diff
@@ -38,9 +38,8 @@ class AdjudicationsSection extends React.Component {
     const { adjudications, loading } = this.state;
     const collectionName = `events/${eventId}/performances/${performanceId}/adjudications`;
     const headings = ['Judge', 'Audio', 'Cumulative Score', 'Awards'];
-    const keys = ['artisticMark', 'audioURL', 'choreoAward', 'cumulativeMark', 'judgeName', 'notes', 'specialAward', 'technicalMark'];
+    const keys = ['artisticMark', 'audioURL', 'choreoAward', 'cumulativeMark', 'notes', 'specialAward', 'tabletID', 'technicalMark'];
     const showAdjudications = Array.isArray(adjudications) && adjudications.length > 0;
-
```
meganniu: Add a line here (?)

src/components/performances/PerformanceForm.jsx
```diff
@@ -198,11 +198,11 @@ PerformanceForm.propTypes = {
     academicLevel: PropTypes.string,
     choreographers: PropTypes.string,
     danceEntry: PropTypes.number,
+    danceSize: PropTypes.number,
```
meganniu: I was under the impression that `danceSize` is a string (like duet, solo etc.)

src/components/performances/PerformanceTableRow.jsx
```diff
@@ -51,11 +51,11 @@ PerformanceTableRow.propTypes = {
     academicLevel: PropTypes.string,
     choreographers: PropTypes.string,
     danceEntry: PropTypes.number,
+    danceSize: PropTypes.number,
```
meganniu: Same question as above: Is `danceSize` a string or a number?

PR #67, opened 2019-03-15
Add Artistic / Technical

YangStephen:
Adds Artistic and Technical Marks onto the UI.

Screen Shot 2019-03-14 at 8 26 30 PM


src/components/adjudications/AdjudicationsSection.jsx
```diff
@@ -37,7 +37,7 @@ class AdjudicationsSection extends React.Component {
     const { match: { params: { eventId, performanceId }}} = this.props;
     const { adjudications, loading } = this.state;
     const collectionName = `events/${eventId}/performances/${performanceId}/adjudications`;
-    const headings = ['Judge', 'Audio', 'Cumulative Score', 'Awards'];
+    const headings = ['Judge', 'Audio', 'Artistic Score', 'Technical Score', 'Cumulative Score', 'Awards'];
```
meganniu: Do we want to keep this consistent with variable names by using "Mark" instead "Score"?

PR #70, opened 2019-05-28
Add Flask Backend

nimwijetunga:
Added Docker for optional local development


backend/requirements.txt
```diff
@@ -0,0 +1,6 @@
+click==6.7
```
imruahmed: can you take a look into using `pipenv` for a virtual environment

backend/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.6
```
imruahmed: can you add docs on how to setup/use docker

backend/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+BUILD Project from root directory (dancefest-web) before starting backend server using the following command:
+```
+$ npm run build
+```
+
+To create and enter the virtual enviroment type the following commands:
+```
+$ pipenv --python 3.5
+$ pipenv install --dev
```
imruahmed: Do these need to be run every time?

backend/README.md
```diff
@@ -0,0 +1,58 @@
...
+$ pipenv shell
+```
+
+Run the following commands to initalize env vars:
+```
+export FLASK_APP=server.py
+export PROJECT_NAME="dancefest-backend"
```
imruahmed: Do these need to be run every time?

imruahmed: Can we add them to the `.env`?

backend/api_routes.py
```diff
@@ -0,0 +1,8 @@
+from flask import Blueprint
+from flask import jsonify
```
imruahmed: unused

backend/README.md
nimwijetunga: I don't usually like putting a .env file in the git repo but if it's fine with you I can add it

PR #75, opened 2019-06-17
Add events update endpoint, seed db

LeozMaxwellJilliams:
- Added events update endpoint
- seeded db with sample data so we can test endpoints
- added `to_dict` to models so we can serialize them


backend/db/models.py
```diff
@@ -11,6 +12,19 @@ class Event(db.Model):
     num_judges = db.Column(db.Integer)
     event_date = db.Column(db.Date)
 
+    def to_dict(self):
```
imruahmed: Can we create a base class which has this method and have all the models inherit?

backend/db/init_db.py
```diff
@@ -15,6 +16,9 @@ def _init_postgres_db(app):
 
     db.session.commit()
 
+    # Seed database with sample data
+    seed()
```
imruahmed: Leave a comment here to take this out once we have real data. This should be in a setup doc, but we'll leave it for now.

backend/db/seeds.py
```diff
@@ -0,0 +1,15 @@
+def seed():
+    from . import db
+    from .models import Event, Performance, Adjudication
+
+    event = Event(event_title="test event")
+    db.session.add(event)
```
imruahmed: Let's use the `Event` class instead of the `db` object. (`Event.create(event_title="test_event")`, etc)

backend/db/seeds.py
```diff
@@ -0,0 +1,15 @@
...
+    db.session.commit()
+
+    for i in range(10):
+        performance = Performance(dance_title='test dance {}'.format(i+1), event_id=event.id)
```
imruahmed: Same with all of these

backend/routes/event_routes.py
```diff
@@ -1,8 +1,33 @@
...
-	return 'Welcome to Dancefest Backend!'
\ No newline at end of file
+	return 'Welcome to Dancefest Backend!'
+
+
+@blueprint.route('/<event_id>', methods=['POST'])
+def update_event(event_id):
+	event = Event.query.get(event_id)
+
+	str_format = '%Y-%m-%d'
```
imruahmed: Create a class of just important constants and import it

backend/routes/event_routes.py
```diff
@@ -1,8 +1,33 @@
...
+
+	event_date = datetime.strptime(request.args['event_date'], str_format)
+	event_title = request.args['event_title']
+	num_judges = request.args['num_judges']
+
+	event.event_date = event_date
```
imruahmed: can inline these things

backend/routes/event_routes.py
```diff
@@ -1,13 +1,34 @@
...
+@blueprint.route('/')
+def main():
+	return 'Welcome to Dancefest Backend!'
+
+
+@blueprint.route('/<event_id>', methods=['POST'])
+def update_event(event_id):
+	event = Event.query.get(event_id)
+
+	event.event_date = datetime.strptime(request.args['event_date'], DATE_FORMAT)
```
imruahmed: Can use the `Event.update` method with the `event_id` specified

PR #77, opened 2019-06-21
Lena/events get

LenaNguyen:
GET events endpoint. Returns all of the events in the db.

### DIFF
--- a/backend/routes/event_routes.py
+++ b/backend/routes/event_routes.py
@@ -39,9 +39,4 @@ def create_event():
 @blueprint.route('/')
 def get_events():
     events = Event.query.all()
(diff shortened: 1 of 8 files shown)


backend/routes/event_routes.py
```diff
@@ -29,3 +31,13 @@ def create_event():
...
+
+@blueprint.route('/')
+def get_events():
+    events = Event.query.all()
+    all_events = {}
+
+    for event in events:
+        all_events[event.id] = event.to_dict()
+
+    return jsonify(all_events)
```
imruahmed: Can replace everything (except the query) with `return jsonify({event.id for event in events})`

PR #81, opened 2019-06-21
Lena/performances get

LenaNguyen:
GET performances given an event id as a parameter. 
Endpoint: /<event_id>/performances

Returns all the performances for a given event.

### DIFF
--- a/backend/routes/performance_routes.py
+++ b/backend/routes/performance_routes.py
@@ -18,27 +18,27 @@ def update_performance(event_id, performance_id):
     '''
     Requires post body in the following format:
     {
-            school: <string>,
-            performers: Array<string>,
-            dance_title: <Int>,
-            dance_style: <string>,
-            dance_entry: <Int>,
-            dance_size: <String>,
-            competition_level: <String>,
-            choreographers: Array<string>,
-            academic_level: <string>	
+                    school: <string>,
+                    performers: Array<string>,
+                    dance_title: <Int>,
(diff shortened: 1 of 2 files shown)


backend/routes/performance_routes.py
```diff
@@ -8,47 +8,53 @@
 
 @blueprint.route('/<event_id>/performances', methods=['POST'])
 def create_performance(event_id):
-	performance_json = request.get_json()
-	new_performance = Performance.create(**performance_json)
-	return jsonify(new_performance.to_dict())
+    performance_json = request.get_json()
```
imruahmed: can you change this back to tabs

backend/routes/performance_routes.py
```diff
@@ -8,47 +8,53 @@
...
 def get_adjudications(performance_id):
-	adjudications = Adjudication.get_by(performance_id=performance_id)
-	return jsonify({adjudication.id: adjudication.to_dict() for adjudication in adjudications})
+    adjudications = Adjudication.get_by(performance_id=performance_id)
+    return jsonify({adjudication.id: adjudication.to_dict() for adjudication in adjudications})
+
+
+@blueprint.route('/<event_id>/performances')
+def get_performances(event_id):
+    all_performances = Performance.get_by(**{"event_id": event_id})
```
imruahmed: i'm not entirely sure, but i think you can simplify this to `Performance.get_by(event_id=event_id)`

PR #86, opened 2019-06-24
Add the post and update end points for adjudications

peiyaaao:


backend/routes/performance_routes.py
```diff
@@ -52,3 +52,19 @@ def update_performance(performance_id):
...
+    adjudication_json['performance_id'] = performance_id
+    new_adjudication = Adjudication.create(**adjudication_json)
+
+    return jsonify(new_adjudication.to_dict())
+
+@blueprint.route('/<performance_id>/adjudications/<adjudications_id>', methods=['POST'])
+def update_adjudication(performance_id, adjudications_id):
+	adjudication = Adjudication.get(adjudications_id)
+	adjudication_json = request.get_json()
+	adjudication.update(**adjudication_json)
```
rogers89: I think imran mentioned that this doesn't update the adjudication local var in place, so
```
adjudication = adjudication.update(...)
```

backend/routes/performance_routes.py
```diff
@@ -52,3 +52,19 @@ def update_performance(performance_id):
...
+@blueprint.route('/<performance_id>/adjudications', methods=['POST'])
+def create_adjudication(performance_id):
+    adjudication_json = request.get_json()
+    adjudication_json['performance_id'] = performance_id
+    new_adjudication = Adjudication.create(**adjudication_json)
+
+    return jsonify(new_adjudication.to_dict())
+
+@blueprint.route('/<performance_id>/adjudications/<adjudications_id>', methods=['POST'])
+def update_adjudication(performance_id, adjudications_id):
```
rogers89: don't need performance id

peiyaaao: @imruahmed  do you think it's sufficient to just remove the performance_id from the route or does it make more sense to create an adjudication_routes.py file with this endpoint inside?

peiyaaao: discussed offline, will move the update endpoint to its own route file but leave the other endpoints in place.

backend/routes/adjudication_routes.py
```diff
@@ -0,0 +1,14 @@
+from flask import Blueprint
+from flask import jsonify, request
+from ..db.models import Adjudication
+
+blueprint = Blueprint('adjudication', __name__, url_prefix='/adjudications')
+
+@blueprint.route('/<adjudications_id>', methods=['POST'])
```
rogers89: pluralized adjudications

PR #91, opened 2019-07-05
Set up structure for client requests

peiyaaao:
This sets up the structure and shows an example get endpoint for adjudications.

### DIFF
--- a/src/api/AdjudicationAPI.js
+++ b/src/api/AdjudicationAPI.js
@@ -1,6 +1,5 @@
 import { get } from './ApiUtils';
-import { getAdjudicationsURL } from './PathConstants';
 
 export const getAdjudications = (id) => {
-    return get(getAdjudicationsURL(id));
+    return get(`performances/${id}/adjudications`);
 };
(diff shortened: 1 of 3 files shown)


src/api/AdjudicationAPI.js
```diff
@@ -0,0 +1,6 @@
+import { get } from './ApiUtils';
+import { getAdjudicationsURL } from './PathConstants';
+
+export const getAdjudications = (id) => {
+    return get(getAdjudicationsURL(id));
```
imruahmed: let's replace `getAdjudicationsURL` with the actual string and remove it from `PathConstants.js`

src/api/ApiUtils.js
```diff
@@ -0,0 +1,27 @@
+import axios from 'axios';
+import { baseURL } from './PathConstants';
+
+const get = (url) => {
```
imruahmed: this is really clean and i like it

src/api/PathConstants.js
```diff
@@ -0,0 +1,2 @@
+export const getAdjudicationsURL = (performanceId) => `performances/${performanceId}/adjudications`;
+export const baseURL = 'http://localhost:5000';
```
imruahmed: your choice: move into `ApiUtils.js` or leave here

PR #92, opened 2019-07-05
Leon/create performance client

LeozMaxwellJilliams:


src/api/performanceApi.js
```diff
@@ -0,0 +1,11 @@
+import axios from 'axios';
+import humps from 'humps';
+import { post } from './ApiUtils';
+
+axios.defaults.headers.post['Access-Control-Allow-Origin'] = '*';
```
imruahmed: remove

src/api/performanceApi.js
```diff
@@ -0,0 +1,11 @@
...
+
+const createPerformance = async (collectionName, data) => {
```
imruahmed: remove `async`

peiyaaao: we can probably do `export const createPerformance.....` 

instead of `export default createPerformance` since we will likely add new apis.

PR #93, opened 2019-07-05
Lena/performances web

LenaNguyen:
endpoint: /events/<eventId>/performances
gets the list of performances for a given event id 
 (you'll need to check the endpoint http://localhost:3000/events/1/performances to view results from the seed data)

### DIFF
--- a/src/components/performances/PerformancesSection.jsx
+++ b/src/components/performances/PerformancesSection.jsx
@@ -1,12 +1,15 @@
+import axios from 'axios';
+import humps from 'humps';
 import React from 'react';
 import PropTypes from 'prop-types';
 import pick from 'lodash/pick';
 
-import db from '../../firebase/firebase';
 import PerformanceDialog from './PerformanceDialog';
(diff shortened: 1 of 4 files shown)


src/components/performances/PerformancesSection.jsx
```diff
@@ -20,25 +23,21 @@ class PerformancesSection extends React.Component {
...
-    this.subscribe();
+	const { match: { params: { eventId }}} = this.props;
+	let performances = [];
+	
+	getPerformances(eventId)
+	.then(response => {
+		Object.values(response.data).forEach(performance => {
+			performances.push(humps.camelizeKeys(performance));
+		});
+		performances = performances.sort((a,b) => Number(a.danceEntry) - Number(b.danceEntry));
```
imruahmed: is there a way to sort this by just casting each `danceEntry` to a `Number`?? i.e. ```performances = performances.sort((performance) => Number(performance.danceEntry));```
If not, then leave it how it is

PR #103, opened 2019-07-15
Leon/mailer

LeozMaxwellJilliams:
updated environment variables


backend/.env.sample
```diff
@@ -1,3 +1,5 @@
 FLASK_APP=server.py
 PROJECT_NAME="dancefest-backend"
 DATABASE_URL="postgres://localhost:5432/dancefest"
+EMAIL_USER="test@example.com"
```
imruahmed: can you change this to something like `replace_me@example.com` and same with password just to let other devs know

PR #106, opened 2019-07-16
Created /feedback and /schoolFeedback routes in App.jsx

ayim:
Creating a PR so that other members have access to the new feedback related routes.


ayim:
Going to create placeholders for the sections so that the app can launch and run.

PR #112, opened 2019-07-19
Add Deployment Support for Heroku

nimwijetunga:
- Refactored codebase to unnest backend
- Created migration file for db to be run on deploy (Procfile)


imruahmed:
The changes look good, but this is gonna cause conflicts for everybody else. I think we should hold off on merging this into apiv2 until the other PRs are merged. Do you see any problems with this?

brandonleung:
note: will not merge. new migration strategy will be based off changes here but this deviates enough that there is too much risk in merging this.

PR #116, opened 2019-07-19
Tablet Routes and Model

rogers89:
Some api infrastructure for tablets needed on the android side.
Allows the android app to:
- Create a new tablet reference using its serial number
- Search tablet existence and ID by serial number
- Search Adjudications by tablet_id


backend/routes/performance_routes.py
```diff
@@ -76,7 +76,8 @@ def get_adjudications_by_performance(performance_ids):
 
 @blueprint.route('/<int:performance_id>/adjudications', methods=['GET'])
 def get_adjudications(performance_id):
-    adjudications = Adjudication.get_by(performance_id=performance_id)
+    # Query by performance id and any additional query parameters provided
+    adjudications = Adjudication.get_by(performance_id=performance_id, **request.args.to_dict())
```
imruahmed: Just making sure you tested this -- I think this works but not entirely sure LOL

imruahmed: Also, can we put a name on `request.args.to_dict()` for readability

PR #118, opened 2019-07-21
Leon/fix client endpoints

LeozMaxwellJilliams:
- fixed create and update performance endpoints on the client side


src/components/performances/PerformanceForm.jsx
```diff
@@ -74,12 +75,14 @@ class PerformanceForm extends React.Component {
   }
 
   handleSubmit = async () => {
-    const { collectionName, formType, performanceId } = this.props;
+    const { collectionName, formType, performanceId, updateData, createData } = this.props;
```
imruahmed: Do we still need `collectionName`? Isn't that old firebase stuff?

src/components/performances/PerformancesSection.jsx
```diff
@@ -46,6 +46,23 @@ class PerformancesSection extends React.Component {
       return res;
     }, []);
 
+  updatePerformances = (data) => {
+    const { performances } = this.state;
+    let newPerformances = performances.map((performance) => {
+      if (data.id === performance.id) return data;
```
imruahmed: Add a comment explaining whats happening here

PR #140, opened 2020-09-15
dependency fix

brandonleung:
hardcoded dep versioning was breaking ability to generate pipfile.lock - updated to latest. works locally.


ChidiNweze (approved):
Approving because Brandon said I should. Hope this helps.

PR #158, opened 2020-12-25
i hope employers see i made a PR on christmas

brandonleung:
- we were calling get school API for every entry in performances (wasteful of network resources)
- turned naive query into join to get school name on first fetch


ChidiNweze (approved):
looks good to me. I hope employers see i told you to go enjoy Christmas with your fam, and stop this madness.

PR #159, opened 2020-12-26
Adjudication admin view

ChidiNweze:
- at the list level show which judges have or have not adjudicated


brandonleung (approved):
woohoo good job

PR #167, opened 2021-01-30
Backend bug fixes + Routes cleanup 

feng-eric:
- Fixed issue where having no adjudication scores would crash the server 
- Added use of service for Performance, Event, Adjudication, and School to decouple code from controller logic to improve maintainability
- Added resource class for Adjudication, Event, and Performance to allow for request body validation
- Commented out code that drops and seeds database every time
- Began adding documentation on route description, request to route, and what it returns
- Added error handling to routes and added status codes to responses in routers: performance, events, adjudication, and school


backend/db/init_db.py
```diff
@@ -7,18 +7,18 @@ def _init_postgres_db(app):
...
 
-    db.session.commit()
+    # db.session.commit()
 
-    # TODO: remove when we have real data
-    # Seed database with sample data
-    seed()
+    # # TODO: remove when we have real data
+    # # Seed database with sample data
+    # seed()
```
mkanoria, line 21: Is this so that the `seed` method is commented out by default?

OustanDing: I think there was some mention somewhere about adding a flag when running the server to specify whether to seed the database. Perhaps we can go with that?

backend/db/models.py
```diff
@@ -130,6 +130,7 @@ class Adjudication(db.Model, BaseMixin):
     performance_id = db.Column(db.Integer, db.ForeignKey('performance.id'))
     nomination_comment = relationship('NominationComment', back_populates="adjudication")
 
+    #TODO: shouldn't this be in create_adjudications in service instead?
```
OustanDing, line 133: It might be cleaner to throw some of this in a service, but I think it might also be okay to leave it here

backend/routes/performance_routes.py
```diff
@@ -1,46 +1,103 @@
 import uuid
 
 from flask import Blueprint
-from flask import jsonify, request
+from flask import jsonify, request, current_app
```
OustanDing, line 4: Combine these two lines?

backend/services/adjudication_service.py
```diff
@@ -0,0 +1,18 @@
...
+    return new_adjudication.to_dict()
+
+def update_adjudication(id, adjudication):
+    Adjudication.query.filter_by(id=id).update(adjudication)
+    update_adjudication = Adjudication.query.get(id)
+    db.session.commit()
+
+    if update_adjudication is None:
+        return update_adjudication
+    return update_adjudication.to_dict()
```
OustanDing, line 18: I think in general we can do like `update_adjudication and update_adjudication.to_dict()` here

backend/services/event_service.py
```diff
@@ -0,0 +1,21 @@
...
+    return new_event.to_dict()
+
+def update_event(id, event):
+    Event.query.filter_by(id=id).update(event)
+    updated_event = Event.query.get(id)
+    db.session.commit()
+
+    if updated_event is None:
+        return updated_event
+    return updated_event.to_dict()
```
OustanDing, line 21: Same as above

OustanDing (approved):
Some small things. BTW do we have an auto Python formatter?

backend/db/init_db.py
feng-eric: Yeah, I think we mentioned that in the future, we wanted to have it seed the DB depending on the environment that we deploy to.

feng-eric:
> Some small things. BTW do we have an auto Python formatter?

Don't think so. @thepenguinco, do you know?

thepenguinco:
> > Some small things. BTW do we have an auto Python formatter?
> 
> Don't think so. @thepenguinco, do you know?



> > Some small things. BTW do we have an auto Python formatter?
> 
> Don't think so. @thepenguinco, do you know?

We don't but we could add one/linter.

PR #172, opened 2021-02-13
Base table component

mkanoria:
[WIP]
The base table seems to be working, need to abstract out some more logic into components

Code is pretty terrible rn, just a PoC


frontend/src/components/Table.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { useTable } from 'react-table';
```
OustanDing, line 2: I think we want to put comments on every import

OustanDing (approved):
LGTM. Just a small comment.

frontend/src/components/Table.js
mkanoria: Thanks, will update!

mkanoria:
Closing since we moved this over on staging

PR #173, opened 2021-02-15
Reusable components

OustanDing:
- Title
- Button
- Text input
- Modal
- Datepicker
- Dropdown

image
image
image


frontend/src/components/Button.js
```diff
@@ -0,0 +1,34 @@
...
+          <Edit />
+        </button>
+      );
+    case 'outlined':
+      return (
+        <button className={`${styles.button} ${styles.button__outlined} ${className}`} {...props}>
+          {children}
+        </button>
+      );
+    default:
```
mkanoria, line 21: Might be a small thing - since the variant is set to oneOf ['outlined', 'contained', 'edit'] maybe 'contained' can be it's own case? Then default could just be a standard button. This might help in the future if say we add more types of buttons. Again, very small nitpick

frontend/src/components/Input.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'; // React
+import PropTypes from 'prop-types'; // PropTypes
+
+import styles from 'styles/components/Input.module.css'; // Component styles
+
+export default function Input({
+  className = '',
+  inputRef = null,
+  icon: Icon,
```
mkanoria, line 9: Did not know about this, nice idea tho!

mkanoria (approved):
LGTM! Added a small comment in the Button component

frontend/src/components/Button.js
OustanDing: I think it's good to keep the `contained` variant as the default case (similar to MUI) since it is already styled, as opposed to a default HTML button. If we want to add more button styles in the future, we can easily just add more cases to this switch.

PR #183, opened 2021-02-21
Feature porting over react components

mkanoria:
Porting over FE components + fixing styling on Login and Events page

Using the standardized Modal component, added `next-images` for svgs 

image

image

image


components/Cards.js
```diff
@@ -45,7 +45,9 @@ function EventCard({ event, isAdmin, openEditModal }) {
           </span>
 
           {/* Event number of performances */}
-          <span>{event.num_performances} Performances</span>
+          <span>
+            {event.num_performances ? event.num_performances : 0} Performances
```
OustanDing, line 49:
```suggestion
            {event.num_performances || 0} Performances
```

OustanDing: Shorter, not a big deal though.

pages/api/events/get.js
```diff
@@ -9,13 +9,15 @@ export default async (req, res) => {
   if (session) {
     // Collect eventID from URL query
     const { id: eventID } = req.query;
-
+    console.log(req.query);
+    console.log(eventID);
     // If eventID is provided
     if (eventID) {
       try {
+        console.log("In here");
```
OustanDing, line 17: We should remove these `console.log`s?

pages/api/events/get.js
```diff
@@ -9,13 +9,15 @@ export default async (req, res) => {
...
         // Collect event with eventID
         let event = await getEventByID(eventID);
-
+        console.log(event);
```
OustanDing, line 20: Same here.

styles/DatePicker.css
```diff
@@ -0,0 +1,164 @@
+/*
```
OustanDing, line 1: Should we make this a SASS file?

OustanDing (approved):
LGTM. A few small comments.

Can we change some of the file names to be singular (eg. `Buttons` --> `Button`) now that we have the reusable components?

pages/api/events/get.js
mkanoria: Good catch LMAO

PR #185, opened 2021-02-21
Adjudication Route

feng-eric:
- Implemented adjudication routes
- Changed audio_url, notes, special award fields in the schema to allow them to be null
- Fixed bug with performance creation


thepenguinco (changes requested):
I think we're going with camelCase for the JS variables (as well as the params and json body fields).

PR #186, opened 2021-02-21
Feedback Sharing

feng-eric:
- Added ability to send feedback email to select schools as well as sending feedback email to all schools
- Added file for AWS SES config (Note: Need to add AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION) to .env
- Added ability to filter performances by schoolIDs
- Added ability to get schools by IDs
- Added returning school data with school contacts
- Changed API routes to plural for consistency

TODO: 
- Format the email body -> to be done after awards are done and a design for what email should look like?
- What should happen if one email fails to send when sending multiple emails?
- Only retrieve performances when all the judging adjudications are submitted.


thepenguinco:
I'll review this in a couple hours when I wake up.

thepenguinco (approved):
LGTM, let's just get another review on this before merging.
I'll add the environment variables into staging and prod environments.

pages/api/feedback/index.js
```diff
@@ -0,0 +1,68 @@
...
+    // Send email for each school
+    schools.forEach(school => {
+      // if there are performances for the school and the school has a contact email
+      if (school.id in performancesMap && 'contacts' in school && school.contacts.length) {
+        // Obtain the first contact email for the school
+        const toEmail = school.contacts[0].email;
+        // Send all performance data for the school
+        transporter.sendMail(
+          {
+            from: 'info@dancefest.dev',
```
thepenguinco, line 48: Let's get this into an environment variable as well.

thepenguinco:
Last thing, can you update the .env.sample file with the new required environment variables.

Thanks!

aws/index.js
```diff
@@ -0,0 +1,14 @@
+const nodemailer = require('nodemailer');
```
OustanDing, line 1: Use `import`?

aws/index.js
```diff
@@ -0,0 +1,14 @@
...
+AWS.config.update({
+  accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+  region: process.env.AWS_REGION,
+});
+
+export const transporter = nodemailer.createTransport({
+  SES: new AWS.SES({
+    apiVersion: '2010-12-01',
+  }),
```
OustanDing, line 13: Won't this create a new object on every import? Is that intentional?

pages/api/feedback/index.js
```diff
@@ -0,0 +1,68 @@
+import { getSession } from 'next-auth/client'; // Session handling
+import { getPerformances } from '../performances/collect';
+import { getSchools } from '../schools/collect';
```
OustanDing, line 3: Are these relative imports okay? @Anish-Agnihotri

pages/api/feedback/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+  if (session) {
+    const { schoolIDs } = req.query;
+
+    // Set filter to obtain schools with schoolIDs passed in. If schoolIDs is not passed in, we want to get all schools.
+    const schoolFilter = schoolIDs ? { id: { in: schoolIDs.split(',').map(i => +i) } } : {};
+    const schools = await getSchools(schoolFilter);
+
+    // If no schools are returned from the filter
+    if (!schools.length) return res.status(400).json({ error: 'No schools to email feedback.' });
```
OustanDing, line 17:
```suggestion
    if (schools.length === 0) return res.status(400).json({ error: 'No schools to email feedback to.' });
```

OustanDing: Might have some linting issues on this line.

pages/api/feedback/index.js
```diff
@@ -0,0 +1,68 @@
...
+      school_id: {
+        in: schools.map(school => school.id),
+      },
+    };
+
+    // Obtain all performances matching the performanceFilter
+    const performances = await getPerformances(performanceFilter);
+
+    // Build a map where we can access all the performances of a school using the school id
+    let performancesMap = {};
```
OustanDing, line 30:
```suggestion
    const performancesMap = {};
```

Not reassigning, so `const` should be fine.

pages/api/feedback/index.js
```diff
@@ -0,0 +1,68 @@
...
+        performancesMap[performance.school_id].push(performance);
+      } else {
+        performancesMap[performance.school_id] = [performance];
+      }
+    });
+
+    // Send email for each school
+    schools.forEach(school => {
+      // if there are performances for the school and the school has a contact email
+      if (school.id in performancesMap && 'contacts' in school && school.contacts.length) {
```
OustanDing, line 42:
```suggestion
      if (school.id in performancesMap && 'contacts' in school && school.contacts.length > 0) {
```

pages/api/performances/collect.js
```diff
@@ -6,21 +6,29 @@ export default async (req, res) => {
...
   // only judges and admins
   // TODO check they have access to this event
   if (session && eventID) {
     // Collect all events from database
-    let performances = await prisma.performance.findMany({
-      where: {
-        event_id: parseInt(eventID),
-      },
-    });
+    let filter = { event_id: parseInt(eventID) };
```
OustanDing, line 16:
```suggestion
    const filter = { event_id: parseInt(eventID) };
```

pages/api/schools/collect.js
```diff
@@ -7,11 +7,26 @@ export default async (req, res) => {
 
   // If session exists (thus, user is authenticated)
   if (session && session.isAdmin) {
-    // Collect all schools
-    let schools = await prisma.schools.findMany();
+    const { schoolIDs } = req.query;
+
+    let filter = {};
```
OustanDing, line 12:
```suggestion
    const filter = {};
```

OustanDing (changes requested):
A couple of comments.

aws/index.js
feng-eric: I followed their documentation so I'm not exactly sure https://nodemailer.com/transports/ses/

feng-eric: Are you saying it should be without the `new`?

OustanDing: Hm, never mind. I don't think it will create a new one. My bad. This is fine.

OustanDing (approved):

pages/api/feedback/index.js
Anish-Agnihotri: Think we should migrate to absolute, and if there isn't one for `@page` and that in `jsconfig.json`.

pages/api/feedback/index.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    // Send email for each school
+    schools.forEach(school => {
+      // if there are performances for the school and the school has a contact email
+      if (school.id in performancesMap && 'contacts' in school && school.contacts.length > 0) {
+        // Obtain the first contact email for the school
+        const toEmail = school.contacts[0].email;
+        // Send all performance data for the school
+        transporter.sendMail(
```
Anish-Agnihotri, line 47: Instead of processing these in a loop, we may want to push the actual requests into an array, and then `Promise.all()` the array execution to parallelize?

Anish-Agnihotri (approved):

pages/api/feedback/index.js
feng-eric: good idea, I made that change

PR #188, opened 2021-02-21
Adjust DatePicker styles

OustanDing:
- Adjust DatePicker styles per Julia's new design specs
- Make outline colour consistent across app


components/DatePicker.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'; // React
+import React, { useRef } from 'react'; // React
```
thepenguinco, line 1: I think Anish has some changes coming in to remove import React, so don't worry about it here.

thepenguinco (approved):

PR #190, opened 2021-02-22
[Feature]: Passing roles via Prisma adapter to session

Anish-Agnihotri:
**Context:**
Removes the hardcoded array of administrator emails and `isAdmin` role handler, in favour of a more dynamic role management system retrieved from the `users` table.

**Implementation:**
Implements a custom `Prisma.Adapter` for `next-auth`, which enables it to pick up on custom fields present in our Prisma schema. Once `next-auth` can detect these fields, we collect the `user.role` from the database and attach to the JWT during the authentication phase. Then, for all subsequent session requests, we strip the role from the JWT and attach to the session.

**Test:**
1. Authenticate with any email, and hit `/api/auth/session`
2. Upgrade your role in the database, and hit `/api/auth/session` (will require a logout)

**Breaking changes:**
Will obviously break all front-end and back-end locations where `isAdmin` is utilized. The very easy mitigation solution for this is to change all checks for `session.isAdmin` to `session.role === 'ADMIN'`. @thepenguinco Let me know if you want me to handle this base case in this PR?


thepenguinco:
Yes please, let's do that for now. We'll need a way to create a master admin user automatically, probably through the seeding scripts.

thepenguinco (changes requested):
Yes please, let's do that quick change on all the routes so we don't break anything. We need a way of getting a master admin into the DB automatically, probably through seeding scripts. Guess let's add seeding to this cycle. Otherwise LGTM.

feng-eric (approved):

prisma/schema.prisma
```diff
@@ -8,14 +8,14 @@ datasource db {
...
-  email          String?   @unique @db.VarChar(255)
-  email_verified DateTime? @db.Timestamptz(6)
-  image          String?
-  created_at     DateTime  @default(now()) @db.Timestamptz(6)
-  updated_at     DateTime  @default(now()) @db.Timestamptz(6)
+  id            Int       @id @default(autoincrement())
+  name          String?   @db.VarChar(255)
+  role          role      @default(USER)
+  email         String?   @unique @db.VarChar(255)
+  emailVerified DateTime? @map("email_verified") @db.Timestamptz(6)
```
OustanDing, line 15: Is there a reason why this is camel case but `created_at` and `update_at` are snake case?

OustanDing (approved):

prisma/schema.prisma
Anish-Agnihotri: Good catch, lol. 

It's a weird `next-auth` condition that expects it to be camelCased. If you change the base property to snake_case it just wont work.

thepenguinco (approved):

PR #195, opened 2021-03-02
[Feature]: Settings & Schools

thepenguinco:
**Requires:**
Schema upgrade required.

**Description:**
Settings and Schools functionality

**Testing:**
Add settings, edit settings, delete settings from the settings page and ensure changes are reflected.
Add schools, edit schools, delete schools from the settings page and ensure changes are reflected.
**TODO**: Test deleting settings that are in use (made easier by DB seeding events/performances)


feng-eric:
In /schools/collect.js and /schools/get.js, we have to remove `include: { contacts: true }`

feng-eric:
In /schools/delete.js, there is a query to delete a contact so that part also needs to be removed

feng-eric (changes requested):
Just a few things that need to be removed in some files since we're removing the contacts table. Other than that, looks good to meeee

pages/api/schools/delete.js
```diff
@@ -12,32 +12,22 @@ export default async (req, res) => {
...
-        res.send(deletedSchoolResult);
-      } else {
-        // internal server error
-        res.status(500).end();
-      }
+
+      // Return deleted performance
+      res.send(deletedSchool);
+    } else {
+      res.status(404).end();
```
feng-eric, line 27: Shouldn't 404 be reserved for if the URL cannot be found? I feel like it should be a 400 error

thepenguinco: 404 for when a resource is not found.  When you delete on subsequent calls, the resource is not available so 404 seems appropriate here.

prisma/schema.sql
```diff
@@ -90,17 +93,4 @@ CREATE TABLE adjudications (
...
-  email VARCHAR(255) NOT NULL,
-  phone VARCHAR(255),
-  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  FOREIGN KEY(school_id) REFERENCES schools(id)
-);
-
-ALTER TABLE settings ADD CONSTRAINT "settings_unique" UNIQUE (type, value);
\ No newline at end of file
+ALTER TABLE settings ADD CONSTRAINT "settings_unique" UNIQUE (type, value);
```
feng-eric, line 96: If we don't care about data when we deploy to prod, we could have this line inside the table right? Since we could create this constraint when the table is created

thepenguinco: @thepenguinco

feng-eric (approved):

PR #196, opened 2021-03-02
[Feature]: Awards Backend Schema

feng-eric:
Changes:
- Implemented award schema (many to many relationship with performance using awardperformance relation schema)
- Added relationship between award and user & award and ajudication
- Implemented CRUD operations for awards
- Added retrieving performances associated with an award and vice versa
- Reformated code structure in performances to improve readability (removed multiple nested if statements)
- Added retrieving school name for each performance
- Added setting userId for created adjudications and awards from session
- Added nominee_count to award performance table and unique constraint (award_id, performance_id)
- Created endpoint to finalize an award and endpoint to nominate a performance for an award
- Improved delete functionality to delete associated adjudications & award performances when deleting an award
- Improved update for performance and award to be able to update provided fields (no longer need to pass all fields)
- Added status to awardperformances table to track which performance is the finalist for the award


thepenguinco:
If you haven't already taken a look here, you can use this to get the joins on relations: https://www.prisma.io/docs/concepts/components/prisma-schema/relations

feng-eric:
> If you haven't already taken a look here, you can use this to get the joins on relations: https://www.prisma.io/docs/concepts/components/prisma-schema/relations

Yup, I did it based on the docs. It seems they use the relation table as the main table for querying

pages/api/adjudications/edit.js
```diff
@@ -15,13 +15,13 @@ export default async (req, res) => {
...
       audioUrl,
       notes,
-      specialAward,
       performanceID,
     } = req.body;
 
     // If all params exist
     if (id && artisticMark && technicalMark && cumulativeMark && performanceID) {
       // Update adjudication
+      // TODO: set user_id from session???
```
feng-eric, line 24: adjudication now should store data about the user who created the adjudication. Should I be getting the id from session or will it be passed from the frontend

thepenguinco: Assume passed for now.  The frontend will/should know the user, user id, user name and user email.

prisma/schema.sql
```diff
@@ -83,10 +83,31 @@ CREATE TABLE adjudications (
...
+  is_finalized BIT NOT NULL,
+  user_id INTEGER NOT NULL,
+  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  FOREIGN KEY(user_id) REFERENCES users(id)
+);
+
+-- Create awards_performances table
+CREATE TABLE awards_performances (
+  id SERIAL PRIMARY KEY NOT NULL,
```
feng-eric, line 107: TODO: add constraint to make (award_id, performance_id) unique

pages/api/adjudications/edit.js
thepenguinco: @feng-eric  actually, scrap that.  Let's use the session in the backend to find out all the user details.  I think that's probably better.

thepenguinco (approved):
Other than what we discussed about last time (addressing some TODOs like adding unique constraint), everything looks fine.

- Let's pull the user ID from the session.  There are multiple ways of doing this, let's just go with attaching the user ID to the session in NextAuth.  We can discuss in meeting.
- We'll need to standardize later error handling 
- Are there any other places we need to do basic error handling (i.e. deleting something and not cleaning up elsewhere will completely break the table schemas)
- Can we get some testing notes/documentation for how to call the specified endpoints maybe?

feng-eric:
> * Are there any other places we need to do basic error handling (i.e. deleting something and not cleaning up elsewhere will completely break the table schemas)
When we delete a performance, we should delete all the performance details (adjudications) associated with it. When we delete an event, we should delete all performances associated with the event, and then all performance details associated with the performance. 
> * Can we get some testing notes/documentation for how to call the specified endpoints maybe?
Yeah, when all changes for awards are done, I'll start on documentation

pages/api/awards/finalize.js
```diff
@@ -0,0 +1,76 @@
...
+
+  // If it does not exist
+  if (!awardPerformance) {
+    return res.status(400).json({
+      error: 'Performance is not nominated for the provided award',
+    });
+  }
+
+  const isFinalized = awardPerformance.awards ? awardPerformance.awards.is_finalized : false;
+  if (isFinalized) {
```
feng-eric, line 43: If we can nominate multiple performances for the same award, I'll remove this

prisma/schema.sql
```diff
@@ -83,11 +86,35 @@ CREATE TABLE adjudications (
...
+  user_id INTEGER NOT NULL,
   created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  FOREIGN KEY(performance_id) REFERENCES performances(id)
+  FOREIGN KEY(user_id) REFERENCES users(id)
+);
+
+-- Create awards_performances table
+CREATE TABLE awards_performances (
+  id SERIAL PRIMARY KEY NOT NULL,
```
feng-eric, line 110: Prisma's doc had this id in addition to the ids of the 2 associated tables in their documentation but it seems a bit redundant

pages/api/awards/collect.js
```diff
@@ -0,0 +1,50 @@
+import prisma from '@prisma/index'; // Prisma client
+import { getSession } from 'next-auth/client'; // Session handling
+
+// Get awards
+export default async (req, res) => {
+  // Collect session from request
+  const session = await getSession({ req });
+
+  //TODO: add functionality of getting awards by performance id
```
feng-eric, line 9: Is this a use case we need

pages/api/awards/collect.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // Collect all awards from database
+  const awards = await getAwards();
+
+  return res.json(awards);
+};
+
+export const getAwards = async () => {
+  // Collect awards with performanceID
+  // TODO: perhaps add flag to include the performances
```
feng-eric, line 24: Do we need to retrieve awards for only certain performances

pages/api/performances/edit.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!id) {
+    return res.status(400).json({
+      error: 'Required performance id not provided',
+    });
+  }
+  console.log(danceTitle);
+
+  const editData = {};
+  if (name) editData.name = name;
```
feng-eric, line 37: Is there a better way of doing this LOL

pages/api/awards/collect.js
thepenguinco: We should get awards by performance for feedback sharing.

pages/api/performances/edit.js
thepenguinco: for edit my impression was the frontend passes in all the data every time they call edit.  no need to support per field 'patching' I think.

pages/api/awards/collect.js
thepenguinco: What do you mean?

pages/api/awards/collect.js
feng-eric: This is already being done in feedback sharing :)

pages/api/awards/collect.js
feng-eric: Currently, when we retrieve an award, we always return the associated performance. Are there cases where we don't want to send this data or do we always want to send it

thepenguinco: Let's always send it.  I see no reason why not to include it.

pages/api/awards/collect.js
thepenguinco: sounds good then we don't need this functionality for now.

prisma/schema.sql
thepenguinco: think it's fine

PR #198, opened 2021-03-05
[Feature]: Performance Details Page

mkanoria:
Building the UI for the Performance Details page

image

image

image

image


mkanoria:
UI is complete 💃💃

pages/performances/[id].js
```diff
@@ -0,0 +1,360 @@
...
+            <h1>{`88`}</h1>
+            <h2>{`CUMULATIVE`}</h2>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+const JudgeFeedback = () => {
```
Comment (mkanoria) [inline pages/performances/[id].js:209]: The Judge Feedback is set to static data for now - can be changed when we add hook up the API

pages/performances/[id].js
```diff
@@ -0,0 +1,360 @@
...
+import styles from '@styles/pages/PerformanceDetails.module.scss';
+
+// Page: Settings
+export default function PerformanceDetails() {
+  const [selectedTab, setSelectedTab] = useState(0);
+  const [showPerformanceSummary, setShowPerformanceSummary] = useState(false);
+  const [feedbackAvailable, setFeedbackAvailable] = useState(true);
+
+  const handleTabClick = () => {
+    setFeedbackAvailable(old => !old); // TODO: Change this!!
```
Comment (OustanDing) [inline pages/performances/[id].js:20]: Can't we just do
```javascript
setFeedbackAvailable(!feedbackAvailable)
```

styles/components/Layout.module.scss
```diff
@@ -86,7 +86,7 @@ Layout: Content
   min-height: calc(100vh - 150px);
 
   .layout__content_sizer {
-    max-width: 1200px;
+    max-width: 1350px;
```
OustanDing, line 89: Does this have any effect on the designs for the other pages?

OustanDing (approved):
Looks very nice! Great work. Just had a small question, otherwise LGTM.

pages/performances/[id].js
Comment (mkanoria) [reply pages/performances/[id].js:20]: Oh yup, good catch. Don't know if it's worth changing though, since this will go away on API hookup. It's just in so that we can toggle b/w the two components (i.e. Empty and JudgeFeedback)

styles/components/Layout.module.scss
mkanoria: It just increases the maximum width of content in each page. I think it does change the layout of the pages, however. I needed it here since the content on this page is supposed to be a little larger than the originally defined 1200px

OustanDing: We should check with Julia on this, just to check if she intended for the width to be different on just one page.

PR #199, opened 2021-03-06
[Feature]: Table Sorting, Filtering, Pagination

OustanDing:
- Added sorting, filtering and pagination to `Table` component
- Dependencies:
  - React Table
  - React Paginate (for pagination component)
- New components:
  - `Pagination` - Page selector for table pagination
  - `Pill` - Pill for filter option

_Note: In the future, we should rework the Performances Page filtering state. Right now, every possible filter has its own state, which requires a bunch of complex object destructuring logic and extra code._


pages/performances/index.js
```diff
@@ -18,62 +22,302 @@ import styles from '@styles/pages/Performances.module.scss'; // Page styles
...
+  const [performanceLevelOptions, setPerformanceLevelOptions] = useState({});
+  const [danceStyleOptions, setDanceStyleOptions] = useState({});
+  const [danceSizeOptions, setDanceSizeOptions] = useState({});
+  const [filters, setFilters] = useState([]);
+  const [pageNumber, setPageNumber] = useState(0);
+  const [pageCount, setPageCount] = useState(0);
+
+  // Get initial filter options
+  useEffect(() => {
+    const getSettings = async () => {
```
mkanoria, line 55: Kinda confused here - is this supposed to be in the `page/performances/index.js` file?

pages/performances/index.js
```diff
@@ -18,62 +22,302 @@ import styles from '@styles/pages/Performances.module.scss'; // Page styles
 // Temp constants
 import data, { columns } from '../../data/mockParticipants';
 
+const PAGE_SIZE = 3; // Rows per page
```
mkanoria, line 25: I'm assuming this changed based on which page we're on? Maybe this was copied over wrong from the setting page?

pages/performances/index.js
```diff
@@ -84,13 +328,59 @@ export default function Performances() {
 }
 
 // Entries Table
-const EntryTable = () => {
-  return <Table columns={columns} data={data} />;
+const EntryTable = props => {
+  const columns = [
```
mkanoria, line 332: Is this different from the `columns` imported on line 23?

mkanoria (commented):
Just a few small comments, but LGTM otherwise!!

pages/performances/index.js
OustanDing: Yes, sorry about the confusing naming. Setting (especially school, dance style and dance level) are used to filter the performances, so they need to be retrieved to populate the filter dropdowns.

pages/performances/index.js
OustanDing: This is a constant that sets the maximum number of rows per page. I set the value to 3 just for testing, but we can set it to whatever is needed.

pages/performances/index.js
OustanDing: Yeah. The `columns` on line 23 is still the mock data which is currently used to populate the Judging table. The `columns` here on line 332 is for the Admin view table. Once the Judging table is done, we won't need the mock columns anymore.

pages/performances/index.js
mkanoria: Ahh so to get the different options for each category, right? Sounds good then

pages/performances/index.js
mkanoria: Right, that makes sense then. We can prob eventually move these column definitions into a separate file in the future

mkanoria:
Just a clarification - this PR covers the new Table functionality + adding that to the Performance page, right? We can do something similar for settings now, I'm assuming?

OustanDing:
> Just a clarification - this PR covers the new Table functionality + adding that to the Performance page, right? We can do something similar for settings now, I'm assuming?

@mkanoria Yeah, exactly. However, IIRC Julia said that we won't be having sorting/filtering/pagination for the settings page?

mkanoria:
> > Just a clarification - this PR covers the new Table functionality + adding that to the Performance page, right? We can do something similar for settings now, I'm assuming?
> 
> @mkanoria Yeah, exactly. However, IIRC Julia said that we won't be having sorting/filtering/pagination for the settings page?

Ohh perfect, let's get this merged in then 💪🏽 🥳

thepenguinco (approved):
LGTM.  Let's see if we can get the data seeding in this cycle as well to test out the whole workflow.

mkanoria (approved):

PR #200, opened 2021-03-06
[Feature]: S3 Media Uploads

Anish-Agnihotri:
Implements the simplest possible workflow for media uploads to S3.

## Rationale
There were two ways to go about solving this problem:
1. Use [nextjs-aws-s3](https://github.com/leerob/nextjs-aws-s3) to implement a route that uses `createPresignedPost`. This mechanism, out-of-the-box, would require us to use CloudFormation via CDK to manage our bucket. It would also mean that we would have to manually manage `multipart/form-data` when posting uploaded content to our back-end routes.
2. **Implemented:** Use [next-s3-upload](https://github.com/ryanto/next-s3-upload) with limited S3 permissions and a hook to manage uploads (handling the form output for us, via a default `APIRoute` export).

## Dependencies
1. Adds a new JS dependency on `next-s3-upload`.
2. Adds an S3 dependency. A new bucket has been created (`s3-dancefest-uploads` in `us-east-1`), along with a new IAM user with permissions to access and edit the bucket. For future reusability, I've created a new policy called `s3-dancefest-upload-policy`—assigning this to any IAM user will give it access to the bucket.

## Usage
1. Add the new environment variables (copy from `.env.sample` into `.env`).
2. Implement, like-so:

```js
// Upload example without front-end loading states, restricting to audio mimetypes, etc.
import { useS3Upload } from 'next-s3-upload';

function uploadSection() {
  const { FileInput, openFileDialog, uploadToS3 } = useS3Upload();

  const handleFileChange = async file => {
    const { url } = await uploadToS3(file);
    console.log(url);
  };

  return (
    <div>
      // This is hidden
      <FileInput onChange={handleFileChange} />

      // Either you prompt for an upload, or our audio recorder would simply use handleFileChange(audio);
      <button onClick={openFileDialog}>Upload File</button>
    </div>
  );
}
```
## Merge blockers
- [X] Heroku environment variables have been updated for both `prod` and `staging` environments.


Anish-Agnihotri:
Aside: By default, [next-s3-upload](https://github.com/ryanto/next-s3-upload) forces the `/api/s3-upload.js` path. I'm not a fan of this, and would much rather we move to something like `/api/media/upload.js` to maintain the folder structure we have going. Perhaps worth a quick PR upstream into the main repository.

thepenguinco (approved):
LGTM, is the route path configurable by default?

Anish-Agnihotri:
> LGTM, is the route path configurable by default?

Nah, I'm going to have to make a PR upstream to the package. I'll write one up tonight.

thepenguinco:
> > LGTM, is the route path configurable by default?
> 
> Nah, I'm going to have to make a PR upstream to the package. I'll write one up tonight.

Hmm probably not the biggest deal in the world then, I'll leave it up to you for whichever option you prefer.

PR #202, opened 2021-03-07
[Bugfix]: Fix Settings page bugs

OustanDing:
Fixes the following bugs:
- Deleting setting value resulting in server error (due to not returning the `res.status(400)` in the API)
- Admin/school modal fields not resetting after closing the modal
- Admin/school still in editing mode even after API request succeeded (which caused new adds to edit the previously edited entry)

New:
- Raise a 400 when creating a new admin with an email that is already being used for another user

### DIFF
--- a/components/Modal.js
+++ b/components/Modal.js
@@ -31,7 +31,10 @@ export default function Modal({
       isOpen={open}
       className={`${styles.modal__container} ${containerClassName}`}
       overlayClassName={styles.modal__overlay}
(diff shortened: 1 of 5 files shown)


thepenguinco (approved):
LGTM.  What is error code P2002?

OustanDing:
@thepenguinco P2002 is the error code returned when violating a unique field constraint. In this case, it would be because a new user is created with an email that another user is already registered with.

PR #207, opened 2021-03-09
[Bugfix]: Revert CSS bug

mkanoria:
- Revert CSS width change
- Updated CSS for performances page to handle width change
- Update `package.json` to run pre-commit hook on `.scss` instead of `.css`

### DIFF
--- a/components/Table.js
+++ b/components/Table.js
@@ -63,7 +63,7 @@ export default function Table({
     setAllFilters(filters);
   }, [filters]);
 
-  const handleClick = row => () => {
+  const handleCellClick = row => () => {
     // Call the prop - onRowClick
     onRowClick(row);
   };
@@ -114,7 +114,7 @@ export default function Table({
             {(paginate ? page : rows).map((row, i) => {
               prepareRow(row);
               return (
-                <tr key={i} {...row.getRowProps()}>
+                <tr key={i} {...row.getRowProps()} onClick={handleRowClick(row)}>
                   {row.cells.map((cell, i) => {
                     return (
                       <td key={i} {...cell.getCellProps()}>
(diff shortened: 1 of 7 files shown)


components/Table.js
```diff
@@ -61,6 +62,11 @@ export default function Table({
     setAllFilters(filters);
   }, [filters]);
 
+  const handleClick = row => () => {
```
OustanDing, line 65: This function name might be a little ambiguous if we also need to handle clicks for other table elements in the future (eg. cells, headers). Perhaps it's worth changing it now, but it's likely not a big deal.

pages/performances/[id].js
```diff
@@ -12,12 +12,12 @@ import styles from '@styles/pages/PerformanceDetails.module.scss';
...
 export default function PerformanceDetails() {
-  const [selectedTab, setSelectedTab] = useState(0);
-  const [showPerformanceSummary, setShowPerformanceSummary] = useState(false);
+  const [selectedTab, setSelectedTab] = useState(-1);
+  const [showPerformanceSummary, setShowPerformanceSummary] = useState(true);
   const [feedbackAvailable, setFeedbackAvailable] = useState(true);
 
   const handleTabClick = () => {
-    setFeedbackAvailable(old => !old); // TODO: Change this!!
+    setFeedbackAvailable(!feedbackAvailable); // TODO: Change this!!
```
Comment (OustanDing) [inline pages/performances/[id].js:20]: Is the `TODO` still necessary here?

components/Table.js
mkanoria: Oh that's a good idea! I'll update it

pages/performances/[id].js
Comment (mkanoria) [reply pages/performances/[id].js:20]: So this `TODO` was for when we hook up the API. Right now, `handleTabClick` just alternates b/w feedback shown and not so that should be changed

PR #216, opened 2021-03-21
[Bugfix]: Fix state management

OustanDing:
- Reset global `event` to `null` upon going to `/`
- Add redirect to `/` on `/performances` and `/awards` if `event` is null


thepenguinco (approved):
LGTM if it's tested!

PR #217, opened 2021-03-21
[Feature]: Moved user_id to awards_performances 

feng-eric:
**Requires:**
- Schema Upgrade

**Description:**
- Moved user_id field to awards_performances table 
- For the response of awards/collect.js, I changed the field name from 'award_performances' to 'performances' inside each awards object

### DIFF
--- a/pages/api/awards/get.js
+++ b/pages/api/awards/get.js
@@ -50,12 +50,15 @@ export const getAward = async filter => {
   // Remove the relation table data
   return {
     ...award,
-    awards_performances: award.awards_performances.map(performance => {
-      return {
-        ...performance.performances,
-        nominee_count: performance.nominee_count,
-        status: performance.status,
-      };
-    }),
+    awards_performances: award.awards_performances.map(
+      ({ performances, nominee_count, status, user_id }) => {
+        return {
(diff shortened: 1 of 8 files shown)


pages/api/awards/get.js
```diff
@@ -50,12 +50,15 @@ export const getAward = async filter => {
...
   return {
     ...award,
-    awards_performances: award.awards_performances.map(performance => {
-      return {
-        ...performance.performances,
-        nominee_count: performance.nominee_count,
-        status: performance.status,
-      };
-    }),
+    awards_performances: award.awards_performances.map(
```
feng-eric, line 53: Is there a good way to rename this field to performances? If I change to awards, the response will have awards_performances from ...award as well. Would deleting the key be the best way? Same question for performances/get.js

thepenguinco: I'm not too sure, we can keep all the information for now.  No extra harm right? @OustanDing

OustanDing: It's fine to keep this as `awards_performances`.

OustanDing (approved):
LGTM!

PR #223, opened 2021-03-26
[Feature]: award schema changes

feng-eric:
- Added eventID into awards schema, `/awards/collect` now requires an eventID in the request body for obtaining the awards for the event. For create, specifying an eventID is now also required.
- In performance and award create + edit, additional validation is added to check that the event exists for the eventID passed in
- Added type (SPECIAL, DANCE_ARTISTRY, SCORE_BASED) into awards schema, removed is_category from awards schema
- Refactored routes to handle the new type enum field in awards
- Added additional validation in `performance/nomination`: Checking that a special award can only be nominated once and ensuring score based awards are not allowed to be nominated
- Fixed existing bug with filtering in `settings/awards`
- Updated seeding to handle the above changes


thepenguinco (approved):
LGTM.

PR #231, opened 2021-04-03
[Feature]: End-to-end audio recording with upload to S3

OustanDing:
- Record judge feedback audio using [React Mic](https://github.com/cleandersonlobo/react-mic), which saves the audio data as a `Blob`. This blob is converted to an mp3 file and is uploaded to S3, after which the adjudication audio url is uploaded with the S3 file link
- The audio file only gets uploaded to S3 if the `Save` button is clicked
- If the audio is not rerecorded in editing mode, it will not be uploaded to S3 and the `Adjudication` DB entry's audio url is left unchanged

If no audio file has been recorded yet:
image

If there exists an audio file already:
image

While recording:
image

### DIFF
--- a/components/AudioPlayer.js
+++ b/components/AudioPlayer.js
@@ -2,9 +2,13 @@ import styles from '@styles/components/AudioPlayer.module.scss';
 
 export default function AudioPlayer({ audioUrl }) {
   return audioUrl ? (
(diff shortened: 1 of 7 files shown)


thepenguinco (approved):
LGTM.

PR #232, opened 2021-04-03
[Improvement]: Change Performances Table routing, move Edit Performances Modal to Performance Details

OustanDing:
- Removed the edit button from the Performances Table. When clicking on a row, the user will be linked to the Performance Details page for the selected performance
- In the Performance Modal, separated the Add functionality from the Edit functionality. Create is now the only option on the Performances Table page. Edit must be done through the Performance Details page

image
image
image

### DIFF
--- a/components/performance-details/EditPerformanceModal.js
+++ b/components/performance-details/EditPerformanceModal.js
@@ -69,6 +69,9 @@ export default function EditPerformanceModal({
       const danceStyleSettings = settings.filter(setting => setting.type === 'STYLE');
       const danceSizeSettings = settings.filter(setting => setting.type === 'DANCE_SIZE');
 
+      // Set the label of each dropdown option to the `value` field of each setting (the setting name in the DB)
+      // Set the selected value (which is sent to API upon save) to the `id` field of each setting (the setting id in the DB)
+      // upon clicking a dropdown option
       const formatOptionsFields = {
         value: 'id',
         label: 'value',
--- a/components/performance-details/EditPerformanceModal.js
+++ b/components/performance-details/EditPerformanceModal.js
@@ -1,27 +1,24 @@
 import React, { useState, useEffect } from 'react'; // React
 import PropTypes from 'prop-types'; // PropTypes
 import axios from 'axios'; // axios
+import Navigation from '@containers/Navigation'; // Navigation state
 
 import Input from '@components/Input'; // Input
 import Modal from '@components/Modal'; // Modal
-import Dropdown from '@components/Dropdown'; // Dropdown
+import Dropdown, { formatDropdownOptions } from '@components/Dropdown'; // Dropdown
 import styles from '@styles/components/performances/PerformanceModal.module.scss'; // Component styles
 
-const EVENT_ID = 1; // Temp event id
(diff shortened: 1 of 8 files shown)


components/performances/PerformancesTable.js
```diff
@@ -47,7 +23,7 @@ export default function PerformancesTable({
       filter: 'matchEnum',
     },
     {
-      Header: 'Level',
+      Header: 'Perf. Lvl',
```
thepenguinco, line 26: Were these abbreviations part of the design?

components/performances/PerformancesTable.js
```diff
@@ -61,11 +37,6 @@ export default function PerformancesTable({
...
       filter: 'matchEnum',
     },
-    {
-      Header: 'Score',
-      accessor: 'score',
-      Cell: ({ value }) => (value !== null ? String(value) : 'N/A'),
-    },
     // Judging view columns
     {
       Header: 'Tech. Score',
```
thepenguinco, line 42: Same here

components/performance-details/EditPerformanceModal.js
```diff
@@ -0,0 +1,284 @@
...
+      const settings = response.data;
+
+      const performanceLevelSettings = settings.filter(
+        setting => setting.type === 'COMPETITION_LEVEL'
+      );
+      const danceStyleSettings = settings.filter(setting => setting.type === 'STYLE');
+      const danceSizeSettings = settings.filter(setting => setting.type === 'DANCE_SIZE');
+
+      const formatOptionsFields = {
+        value: 'id',
```
thepenguinco, line 73: I remember we had a bug surrounding what value and label were, can we put a comment somewhere around these so its clear what exactly id, value and label are referring to (especially since we have 2 'values' here.

components/performance-details/EditPerformanceModal.js
```diff
@@ -0,0 +1,284 @@
...
+          choreographers: choreographersString
+            .split(',')
+            .map(choreographer => choreographer.trim()),
+          competitionLevel: competitionLevel.label,
+          competitionLevelID: competitionLevel.value,
+          danceSize: danceSize.label,
+          danceSizeID: danceSize.value,
+          danceStyle: danceStyle.label,
+          danceStyleID: danceStyle.value,
+          danceTitle,
```
thepenguinco, line 127: Careful with the name and danceTitle here.  Can you double to check to make sure we're actually cleaning this up on the frontend?

thepenguinco (approved):
Small changes, please see comments.  Otherwise LGTM.

components/performance-details/EditPerformanceModal.js
thepenguinco: We agreed on just using danceTitle.  I'll be pushing out the backend changes tonight for this.

components/performances/PerformancesTable.js
OustanDing: Yup

components/performances/PerformancesTable.js
OustanDing: Also in the design

PR #233, opened 2021-04-03
[Bugfix]: Automatically calculate cumulative score

OustanDing:
- A user was able to manually input the cumulative score for a performance
- Disallow the cumulative score from being manually edited - calculate it automatically by averaging the technical and artistic scores


thepenguinco (approved):
LGTM

PR #234, opened 2021-04-04
[Bug]: Performance count for each event

feng-eric:
- Removed performance_count field in event that was not being used
- Changed query in `event/collect` and `event/get` to obtain associated performances to obtain number of performances for the event
- For frontend: remove uses of the `performance_count` field and use `performances` instead to display the number of performances


mkanoria (approved):
LGTM! Might break something on the events page since we're looking for `num_performances` but that can be fixed easily

feng-eric:
> LGTM! Might break something on the events page since we're looking for `num_performances` but that can be fixed easily

@mkanoria I can make that fix in this PR too!

mkanoria:
Oh awesome!! Change looks good to merge!

PR #238, opened 2021-04-06
[Feature]: Judge View

mkanoria:
- Add judge view to screens across the app
  - disable settings from navbar
  - update awards and performances page to show data specific to judge
  - disable adding or editing performances, deleting awards 
- Fix small dance title bug in the awards page

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "@babel/runtime": "^7.13.10",
     "@cleandersonlobo/react-mic": "^1.2.0",
     "@prisma/client": "2.20.1",
     "axios": "^0.21.1",
--- a/pages/api/schools/collect.js
+++ b/pages/api/schools/collect.js
@@ -6,7 +6,7 @@ export default async (req, res) => {
   const session = await getSession({ req });
 
   // If session exists (thus, user is authenticated)
-  if (session && session.role === 'ADMIN') {
+  if (session && (session.role === 'ADMIN' || session.role === 'JUDGE')) {
     const { schoolIDs } = req.query;
 
     const filter = {};
--- a/pages/api/user/collect.js
+++ b/pages/api/user/collect.js
@@ -6,7 +6,7 @@ export default async (req, res) => {
   const session = await getSession({ req });
 
   // If session exists (thus, user is authenticated)
-  if (session && session.role === 'ADMIN') {
+  if (session && (session.role === 'ADMIN' || session.role === 'JUDGE')) {
     const { role } = req.query;
 
     const filter = {};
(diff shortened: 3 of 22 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "@babel/runtime": "^7.13.10",
```
mkanoria, line 18: The project was not compiling for me locally without adding this, but can discuss if this was an issue only on my setup

pages/api/schools/collect.js
```diff
@@ -6,7 +6,7 @@ export default async (req, res) => {
   const session = await getSession({ req });
 
   // If session exists (thus, user is authenticated)
-  if (session && session.role === 'ADMIN') {
+  if (session && (session.role === 'ADMIN' || session.role === 'JUDGE')) {
```
mkanoria, line 9: Without this the API returns a `401` for judge roles - we should return the data but filter it from the user on the FE

pages/api/user/collect.js
```diff
@@ -6,7 +6,7 @@ export default async (req, res) => {
   const session = await getSession({ req });
 
   // If session exists (thus, user is authenticated)
-  if (session && session.role === 'ADMIN') {
+  if (session && (session.role === 'ADMIN' || session.role === 'JUDGE')) {
```
mkanoria, line 9: Same as above

package.json
thepenguinco: I guess this is fine.  Very strange behaviour though.

pages/api/schools/collect.js
thepenguinco: Why do the judges need to be able to call api/schools/collect, they never need to see all the schools (just the ones associated with the appropriate performances).

pages/api/user/collect.js
thepenguinco: Why do the judges need to be able to call api/user/collect?  They never need to access the settings page.

thepenguinco (approved):
Few comments.  Otherwise LGTM.

pages/api/schools/collect.js
mkanoria: Looked into it, this is used to populate the filter options for the [performances page](https://github.com/uwblueprint/dancefest-web/blob/staging/pages/performances/index.js#L114)

pages/api/user/collect.js
mkanoria: This is to populate the 'judges' field of each event card iirc in [index.js](https://github.com/uwblueprint/dancefest-web/blob/staging/pages/performances/index.js#L114)

thepenguinco:
Okay got it, that's fine then.

PR #239, opened 2021-04-07
[Bugfix]: Disable button while in loading state

OustanDing:
Previously, buttons were still clickable after submitting a request using them. This caused event to be fired more than once, which was unideal, especially when it came to DB-modifying actions.

- Disabled all buttons that fire API requests while in loading states


thepenguinco (approved):
LGTM.

PR #240, opened 2021-04-09
[Bugfix]: Judge Views

mkanoria:
**Bugfix**
- Hide Add Award button for judge view

**Feature**
- Add delete award button for score based awards
- Add audio player component to awards details

Should have broken this into multiple PR's but they're all small features so figured I'd get them in together. Can make separate PR's if needed

### DIFF
--- a/components/awards/ScoreBasedAwards.js
+++ b/components/awards/ScoreBasedAwards.js
@@ -5,6 +5,7 @@ import Layout from '@components/Layout'; // Layout wrapper
 
 import Navigation from '@containers/Navigation'; // Navigation state
 import Loader from 'react-loader-spinner'; // Spinning loader
+import Button from '@components/Button'; // Button
 import Title from '@components/Title'; // Title
 import Input from '@components/Input'; // Input
 import Tabs from '@components/Tabs'; // Tabs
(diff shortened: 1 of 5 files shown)


components/awards/ScoreBasedAwards.js
```diff
@@ -113,6 +117,23 @@ export default function Performances({ award }) {
     }
   }
 
+  async function deleteAward() {
+    setLoading(true);
+    try {
+      await axios({
+        method: 'PUT',
```
thepenguinco, line 124: POST?

thepenguinco (approved):
LGTM.

components/awards/ScoreBasedAwards.js
mkanoria: I think we call `PUT` everywhere else, but I agree, it should be `POST`

mkanoria:
Need to do some more testing of the Audio Player change locally, will merge after if everything looks good

pages/awards/[id].js
```diff
@@ -367,12 +367,16 @@ const IndividualFeedback = ({ feedback }) => {
...
+      <div className={styles.judge__feedback_audio_player_wrapper}>
+        <FeedbackAudio
+          audioUrl={feedback.audio_url}
+          edit={false}
+          setRecordingChanged={() => {}}
+          recording={null}
+          setRecording={() => {}}
+          recordedBlob={null}
+          setRecordedBlob={() => {}}
+        />
```
Comment (OustanDing) [inline pages/awards/[id].js:379]: If we're only allowing playback here, please use the `AudioPlayer` component.

Comment (mkanoria) [reply pages/awards/[id].js:379]: Good catch, will update!

PR #242, opened 2021-04-11
[Feature]: backend validations for editing and deleting adjudications, performances, awards

feng-eric:
### Deleting Adjudications (Not in frontend ATM)
- Check that adjudication exists
- If the performance associated with the adjudication is finalized, prevent deleting
- Delete all nominations with the same user_id as the user_iud in adjudication for the performance

### Edit Adjudications
- Check that adjudication exists
- If the performance associated with the adjudication is finalized, prevent editing -> Tested

### Delete Awards
- Check that award exists
- If the award is finalized, prevent deletion -> Tested

### Edit Awards (Not in frontend ATM)
- Removed ability to set is_finalized through edit endpoint
- Check that award exists
- If the award is finalized, prevent editing
- If the award has nominations, prevent editing

### Delete Performance (Not in frontend ATM)
- Check that performance exists
- If the performance has adjudications, prevent 
- If the performance is finalized, prevent deletion 
- Removed deleting award_performance records in cascading delete

### Edit Performance
- Check that performance exists 
- If the performance has adjudications, prevent editing -> Tested
- If the performance is finalized, prevent editing -> Tested

### Nominate
- If the performance is already finalized, we stop the nominations from being edited or more nominations to be added -> Tested. 
- If one of the awards attempted to be nominated is already finalized, we prevent the nominations from occurring completely. -> Tested

### DIFF
--- a/pages/api/performances/delete.js
+++ b/pages/api/performances/delete.js
@@ -44,8 +44,8 @@ export default async (req, res) => {
 
   // If the performance is finalized, we do not allow deletion
   if (performance.awards_performances && performance.awards_performances.length !== 0) {
-    for (const nomination in performance.awards_performances) {
-      if (nomination.status === 'FINALIST') {
+    for (const index in performance.awards_performances) {
+      if (performance.awards_performances[index].status === 'FINALIST') {
         return res.status(400).json({
           error: 'Performance cannot be deleted as it is finalized',
         });
--- a/pages/api/performances/delete.js
+++ b/pages/api/performances/delete.js
@@ -44,8 +44,8 @@ export default async (req, res) => {
 
   // If the performance is finalized, we do not allow deletion
   if (performance.awards_performances && performance.awards_performances.length !== 0) {
-    for (const index in performance.awards_performances) {
-      if (performance.awards_performances[index].status === 'FINALIST') {
+    for (const nomination of performance.awards_performances) {
+      if (nomination.status === 'FINALIST') {
         return res.status(400).json({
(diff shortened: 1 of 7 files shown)


thepenguinco (approved):
LGTM.  Were you able to test out some of the changes via the frontend as well?

feng-eric:
> LGTM. Were you able to test out some of the changes via the frontend as well?

Not yet. I'm planning to do that first so I won't merge it in until I do that tomorrow

pages/api/performances/delete.js
```diff
@@ -18,13 +18,42 @@ export default async (req, res) => {
...
+  if (performance.adjudications && performance.adjudications.length !== 0) {
+    return res.status(400).json({
+      error: 'Performance cannot be edited as it has an adjudication',
+    });
+  }
+
+  // If the performance is finalized, we do not allow deletion
+  if (performance.awards_performances && performance.awards_performances.length !== 0) {
+    for (const index in performance.awards_performances) {
+      if (performance.awards_performances[index].status === 'FINALIST') {
```
OustanDing, line 48: `in` is for iterating through object keys. If this is an array, you need to use `of`. Same applies below.

feng-eric: Oh oops I mixed it up since the changes I made in awards was iterating through object keys. Fixed.

thepenguinco (approved):
LGTM!  Likely most of the incoming bugs I foresee from the dancefest team will be frontend issues.

PR #246, opened 2021-04-13
[Feature]: Per environment seeding for production

thepenguinco:
**Description:** 
- Separate prod and dev seeding via environment variable (see .env sample)
- Bump version number

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -5,6 +5,9 @@ NA_EMAIL_SERVER="smtp://yourmailserver.com"
 NA_EMAIL_FROM="login@dancefest.dev"
 NA_JWT_SECRET="some_random_secret"
 
+# Seed ("production" for production seeding, anything else for dev seeding)
+SEED=""
+
 # Database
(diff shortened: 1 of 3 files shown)


.env.sample
```diff
@@ -5,6 +5,9 @@ NA_EMAIL_SERVER="smtp://yourmailserver.com"
 NA_EMAIL_FROM="login@dancefest.dev"
 NA_JWT_SECRET="some_random_secret"
 
+# Seed ("production" for production seeding, anything else for dev seeding)
+SEED=""
```
feng-eric, line 9: Not super important but if we will have different variables somewhere else depending on the environment, you can have it be `ENVIRONMENT` instead of `SEED`. For example, the use case where the recipient email for feedback sharing is different depending on if it's production vs. staging?

feng-eric: Unless we want to seed the staging database for schools with our emails lol (currently I think they're fake emails)

PR #251, opened 2021-04-14
[Bug]: Fixed backend crashing because multiple responses were being sent

feng-eric:
- Changed res.send() to res.json() whenever appropriate
- Added 'return' infront of res just in case


mkanoria (approved):
LGTM! Thanks for finding the fix! :shipit:

thepenguinco (approved):
LGTM.  Good catch.

PR #252, opened 2021-04-15
[Feature]: Validation for score card input

OustanDing:
Add validation for score card input:
- Disallow non-integer inputs
- Disallow keys other than numeric keys
- Disallow scores < 0 or scores > 100
- Still allow important keys (ie. arrow keys, backspace/delete, etc)

Note: The implementation was very hacky due to race conditions, and a lot of the keypress handling logic is hard-coded because of the lack of support in HTML when using `<input type="number" />`. Copy-pasting may get around some of the validations, but that should be very infrequent relative to manual input. I could patch that, but it would make the code extremely complex/hard to maintain.


feng-eric (approved):
Looks good! How significant is the race condition issue? Is it very likely to occur

OustanDing:
@feng-eric Good question. The race condition occurred very often, as the input's set state function was racing against the manual input, I believe. The solution I implemented in this PR uses `onKeyDown`, which guarantees that the logic in that handler will take place before `onChange`.

thepenguinco (approved):
Sanity check on staging.  Otherwise LGTM.

mkanoria (approved):

feng-eric (approved):

PR #255, opened 2021-04-17
[Bugfix]: Fix Event modal submit button disable state

OustanDing:
Event modal Submit button was disabled on the first click if the input was invalid. However, on subsequent clicks, it remained disabled.

In this PR:
- Fixed the above bug
- Cleaned up Modal frontend rendering. Instead of conditionally rendering the content inside the modal, we should conditionally render the entire modal. While this is less ideal performance-wise, it is a lot easier to maintain and debug for developers
- Fixed modal styling to scroll when the judges overflow

Fixed Event modal Submit button to be enabled when all of the following are satisfied:
- Dance title is a non-empty string
- Date is non-empty
- There is at least one judge
- At least one selected judge is non-null (user has selected one of the possible judges)


feng-eric (approved):
Looks good! @thepenguinco, wanted to confirm if we want to clean up the db schema, or should we leave it the way it is? I went to look at certain fields for events and noticed we have a location field that isn't being used and also fields like event_date and judges can be null (however, it's prevented from being null by both frontend + backend).

thepenguinco:
> Looks good! @thepenguinco, wanted to confirm if we want to clean up the db schema, or should we leave it the way it is? I went to look at certain fields for events and noticed we have a location field that isn't being used and also fields like event_date and judges can be null (however, it's prevented from being null by both frontend + backend).

Rather not have schema changes now, but let's make a note of it for the future.  Thanks!

thepenguinco (approved):

PR #257, opened 2021-04-18
[Feature]: None option in the Add Award modal

mkanoria:
- Add None option to the modal
 - Remove unused ID from scorebased award table + fix missing performance name bug


thepenguinco (approved):
LGTM.  Sanity checks and then let's ship it :D!

PR #259, opened 2021-04-19
[Feature]: Added filtering for performances for score based awards

feng-eric:
- Added filtering for performances on score based awards page so that only performances eligible for the score based award are shown.

From the award_categories field of award, do we know which one corresponds to competition_level_id, dance_size_id, or dance_style_id? Right now, I have it send all the settingIDs in a list. If it's possible to know what each one corresponds to, I believe it would make the backend a bit cleaner since right now, for each of the 3 fields in performances, we're checking if the id is in the settingIDs array that's passed in.

### DIFF
--- a/components/awards/ScoreBasedAwards.js
+++ b/components/awards/ScoreBasedAwards.js
@@ -56,6 +56,7 @@ export default function Performances({ award, session }) {
   const getPerformances = async () => {
     setLoading(true);
 
+    // NOTE: If we know the specific award category, we can pass in 3 params (competition_level_id, dance_size_id, dance_level_id)
     let performanceUrl = `/api/performances/collect?eventID=${event.id}`;
     if (award.awards_categories && award.awards_categories.length !== 0) {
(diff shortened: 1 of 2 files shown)


components/awards/ScoreBasedAwards.js
```diff
@@ -56,10 +56,15 @@ export default function Performances({ award, session }) {
   const getPerformances = async () => {
     setLoading(true);
 
+    let performanceUrl = `/api/performances/collect?eventID=${event.id}`;
+    if (award.awards_categories && award.awards_categories.length !== 0) {
+      performanceUrl += `&settingIDs=${award.awards_categories.join(',')}`;
```
feng-eric, line 61: If it's possible to know which element in the awards_category array corresponds to, I can do 3 query parameters ex. competition_level, dance_size, dance_style. Is that possible to know?

thepenguinco: There is a setting type which you can look up with a corresponding SettingType enum.

thepenguinco (approved):
LGTM.

PR #268, opened 2021-04-23
[Bugfix]: Fix judges seeing other judges' feedback

OustanDing:
When a judge was filling out the feedback for a performance, they were possibly shown the feedback given by another judge. This bug was due to the logic serving the adjudication not accounting for a judge.


thepenguinco (approved):
LGTM.

PR #269, opened 2021-04-23
[Improvement]: Reorder scores

OustanDing:
Some places in the app showed artistic score before technical score, but we want to show technical score first.


thepenguinco (approved):
LGTM.

PR #270, opened 2021-04-25
[Bugfix]: Fix score based filtering, finalize bug fix

mkanoria:
- Fix pre-filtering for score based awards
  - Only check setting ID by the setting category they belong in
- Fix finalize bug for score based awards
  - For awards_performances, only check if a performance is nominated for the specific award it's being nominated for

### DIFF
--- a/components/Table.js
+++ b/components/Table.js
@@ -42,6 +42,15 @@ export default function Table({
             return rowValue !== undefined ? filterValues.includes(String(rowValue)) : true;
           });
         },
+        matchCategory: (rows, id, filterValues) => {
+          // Match an array to another array
+          return rows.filter(row => {
+            const rowValue = row.values[id];
+            return rowValue !== undefined
+              ? filterValues.some(v => rowValue.indexOf(parseInt(v)) >= 0)
+              : true;
+          });
+        },
       },
       initialState: {
         sortBy: initialSort,
--- a/pages/api/performances/collect.js
+++ b/pages/api/performances/collect.js
@@ -34,13 +34,14 @@ export default async (req, res) => {
 
     for (const settingID of settingIDArray) {
       const setting = await getSetting(settingID);
-
-      if (setting.type === 'COMPETITION_LEVEL') {
-        filter.competition_level_id = settingID;
-      } else if (setting.type === 'STYLE') {
-        filter.dance_style_id = settingID;
-      } else if (setting.type === 'DANCE_SIZE') {
-        filter.dance_size_id = settingID;
+      if (setting) {
+        if (setting.type === 'COMPETITION_LEVEL') {
+          filter.competition_level_id = settingID;
+        } else if (setting.type === 'STYLE') {
+          filter.dance_style_id = settingID;
+        } else if (setting.type === 'DANCE_SIZE') {
+          filter.dance_size_id = settingID;
+        }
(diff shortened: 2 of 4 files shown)


pages/api/performances/collect.js
```diff
@@ -102,3 +103,13 @@ export const getPerformances = async filter => {
     }
   );
 };
+
+const getSetting = async id => {
+  let setting = await prisma.setting.findUnique({
```
feng-eric, line 108: Can prob change to use `const`

pages/api/performances/collect.js
```diff
@@ -102,3 +103,13 @@ export const getPerformances = async filter => {
...
+    where: {
+      id: parseInt(id),
+    },
+  });
+  if (!setting) return {};
```
feng-eric, line 113: Did you test when it doesn't find a setting with the id?

feng-eric (approved):
Left some comments but otherwise, it looks good to me!

pages/api/performances/collect.js
mkanoria: Don't anticipate this happening but I added a check. Thanks for catching this!

pages/api/performances/collect.js
```diff
@@ -31,17 +31,19 @@ export default async (req, res) => {
...
+        if (setting.type === 'COMPETITION_LEVEL') {
+          filter.competition_level_id = settingID;
+        } else if (setting.type === 'STYLE') {
+          filter.dance_style_id = settingID;
+        } else if (setting.type === 'DANCE_SIZE') {
+          filter.dance_size_id = settingID;
+        }
+      }
+    }
   }
```
OustanDing, line 47: This is an N+1 query problem. API performance shouldn't be too much worse but with the `Performances` page being the most important page, we should consider optimizing this.

feng-eric: What could be done instead is that in awards/collect and awards/get, we can get the setting type along with the setting ids. Then, we wouldn't have to query to find the type in this for loop

components/Table.js
```diff
@@ -42,6 +42,15 @@ export default function Table({
             return rowValue !== undefined ? filterValues.includes(String(rowValue)) : true;
           });
         },
+        matchCategory: (rows, id, filterValues) => {
```
mkanoria, line 45: Added a new filter type since `matchEnum` checked from an array to a single value, whereas here we can have multiple filters but also the categories column is an array so have to compare it slightly different

pages/awards/index.js
```diff
@@ -26,18 +27,24 @@ import styles from '@styles/pages/Awards.module.scss'; // Page styles
 import useSnackbar from '@utils/useSnackbar'; // Snackbar
 
 const PAGE_SIZE = 20; // Rows per page
+const AWARDS_TABLE_HIDDEN_COLUMN = ['categories'];
```
mkanoria, line 30: Have to add the column to filter in the actual table otherwise React Table does not run filter on it. So I decided to just add it as a column and hide it

pages/api/performances/collect.js
OustanDing: Decided to leave this out for now cc @mkanoria

OustanDing (approved):

PR #273, opened 2021-04-26
[Bugfix]: Small fixes

OustanDing:
- Make `fixed` property for `Dropdown` component optional; only apply to Event modal
- Fix hard-coded "Event title"
- Fix award type showing enum


thepenguinco (approved):
LGTM.

